PR #2, opened 2020-03-17
Jaicp async channels support

Denire:
Changes:

1. Add support for JAICP async channels using long polling or webhook connection;
2. Create package with JAICP Telephony examples
3. Remove obsolete references to projectId. AccessToken will be used instead.


examples/jaicp-telephony/src/main/kotlin/com/justai/jaicf/examples/jaicptelephony/TelephonyBotModule.kt
```diff
@@ -1,12 +1,12 @@
...
+package com.justai.jaicf.examples.jaicptelephony
 
 import com.justai.jaicf.channel.jaicp.channels.TelephonyEvents
 import com.justai.jaicf.channel.jaicp.dto.telephony
 import com.justai.jaicf.channel.jaicp.reactions.telephony
 import com.justai.jaicf.helpers.logging.WithLogger
 import com.justai.jaicf.model.scenario.Scenario
 
-class CitiesGameBotModule : Scenario(), WithLogger {
+class TelephonyBotModule : Scenario(), WithLogger {
```
morfeusys: Why not to use `object` instead of class?

Denire: Yep, thanks for noticing.

PR #14, opened 2020-04-16
add basic tests for jaicp channels

Denire:
Add basic tests for jaicp channels using webhook connector


Denire:
will be reincarnated afterwards

PR #19, opened 2020-06-21
Initial support of updated Alice API

Pazus:
Need some assistance in the way such changes are tested in the framework.

### DIFF
--- a/channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
+++ b/channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
@@ -3,7 +3,7 @@ package com.justai.jaicf.channel.yandexalice.api
 import com.justai.jaicf.api.BotRequest
 import com.justai.jaicf.api.BotRequestType
 import com.justai.jaicf.channel.yandexalice.AliceEvent
-import com.justai.jaicf.channel.yandexalice.api.model.IntentName
+import com.justai.jaicf.channel.yandexalice.api.model.AliceIntent
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonElement
@@ -18,7 +18,6 @@ data class AliceBotRequest(
     val meta: Meta? = null,
     val session: Session,
     val request: Request? = null,
-    @SerialName("state")
     val state: State? = null,
     @SerialName("account_linking_complete_event")
     val accountLinkingCompleteEvent: JsonObject? = null
@@ -65,10 +64,12 @@ data class Session(
     @SerialName("skill_id")
     val skillId: String,
     val application: Application,
-    val user: User?
+    val user: User? = null
 ) {
+    // see: https://yandex.ru/dev/dialogs/alice/doc/protocol-docpage/
+    // application.applicationId contains the same value `user_id` contained
     @Deprecated(
-        message = "use application.applicationId instead",
+        message = "use application.applicationId instead as recommended in documentation",
         replaceWith = ReplaceWith(expression = "application.applicationId")
     )
     val userId: String
@@ -92,10 +93,8 @@ data class Application(
 @Serializable
 data class State(
     val session: JsonObject? = null,
-    val user: Map<String, JsonElement>? = null
+    val user: Map<String, JsonElement> = emptyMap()
 )
-
-typealias SlotName = String
 
 @Serializable
 data class Request(
@@ -111,7 +110,10 @@ data class Request(
     data class Nlu(
         val tokens: List<String>,
         val entities: List<Entity>,
-        val intents: Map<IntentName, Intent>
+        /**
+         * @see AliceIntent for predefined
+         * */
+        val intents: Map<String, Intent>
     ) {
         @Serializable
         data class Entity(
@@ -128,16 +130,13 @@ data class Request(
 
         @Serializable
         data class Intent(
-            val slots: Map<SlotName, Slot> = emptyMap()
+            val slots: Map<String, Slot> = emptyMap()
         )
 
         @Serializable
         data class Slot(
-            @SerialName("type")
             val type: String,
-            @SerialName("value")
             val value: JsonElement,
-            @SerialName("tokens")
             val tokens: Tokens?
         )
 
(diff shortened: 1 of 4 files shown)


channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
```diff
@@ -17,10 +18,12 @@ data class AliceBotRequest(
     val meta: Meta? = null,
     val session: Session,
     val request: Request? = null,
+    @SerialName("state")
```
morfeusys, line 21: Why do we need this annotation here once the field has the same name?

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
```diff
@@ -17,10 +18,12 @@ data class AliceBotRequest(
...
+    val state: State? = null,
     @SerialName("account_linking_complete_event")
     val accountLinkingCompleteEvent: JsonObject? = null
 ): BotRequest {
-    override val clientId = session.userId
+    override val clientId = session.application.applicationId
```
morfeusys, line 26: `clientId` must be unique for every user of the skill, not for a while application

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
```diff
@@ -59,12 +62,41 @@ data class Session(
     val messageId: Int,
     @SerialName("session_id")
     val sessionId: String,
+    @SerialName("skill_id")
+    val skillId: String,
+    val application: Application,
+    val user: User?
+) {
+    @Deprecated(
```
morfeusys, line 70: Regarding the actual documentation this field is not deprecated

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
```diff
@@ -59,12 +62,41 @@ data class Session(
     val messageId: Int,
     @SerialName("session_id")
     val sessionId: String,
+    @SerialName("skill_id")
+    val skillId: String,
+    val application: Application,
+    val user: User?
```
morfeusys, line 68: Should be `null` be default for the deserialization purposes?

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
```diff
@@ -59,12 +62,41 @@ data class Session(
...
+@Serializable
+data class Application(
+    @SerialName("application_id")
+    val applicationId: String
+)
+
+@Serializable
+data class State(
+    val session: JsonObject? = null,
+    val user: Map<String, JsonElement>? = null
```
morfeusys, line 95: Would be better to have an empty map initialised here by default?

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
```diff
@@ -92,5 +125,26 @@ data class Request(
...
         }
+
+        @Serializable
+        data class Intent(
+            val slots: Map<SlotName, Slot> = emptyMap()
+        )
+
+        @Serializable
+        data class Slot(
+            @SerialName("type")
```
morfeusys, line 136: `@SerialName` is redundant here?

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotResponse.kt
```diff
@@ -5,18 +5,22 @@ import com.justai.jaicf.channel.yandexalice.api.model.Button
...
 data class AliceBotResponse(
     var response: Response?,
     @SerialName("start_account_linking")
     var startAccountLinking: JsonObject? = null,
-    val session: Session,
-    val version: String
+    val version: String,
+    @Deprecated(message = "May be omitted")
+    val session: Session? = null,
+    val userStateUpdate: MutableMap<String, JsonElement?> = mutableMapOf(),
```
morfeusys, line 19: It would be great to have some logic somewhere that propagates the request's session_state to this map.
Also I'd change the type of this field to `JsonObject` if possible.

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/model/AliceIntent.kt
```diff
@@ -0,0 +1,10 @@
+package com.justai.jaicf.channel.yandexalice.api.model
+
+typealias IntentName = String
```
morfeusys, line 3: I'm not sure we heed this typealias..

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
Pazus: https://yandex.ru/dev/dialogs/alice/doc/protocol-docpage/ Look at user_id field
```
Осторожно. Свойство перестает поддерживаться — вместо него следует использовать новое, полностью аналогичное свойство session.application.application_id.
```

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotResponse.kt
Pazus: We disagree on that. The skills is going to be stateless as original and new states are provided by the API. If the framework provides logic to copy original state to the new one skill will loop forever in the same state. The new state even if its a copy of the original should be set explicitly by the developer. This is the reasoning behind having it in a separate field in the API.

Pazus: User state is incremental, meaning that the skill returns only increment to the user_state in response as it's shared among all surfaces of the user and there might be a race in changing it. That's why in the response MutableMap if used.
By design a developer has to pass `null` value associated with a key to remove the key's entry from the user state.
Details are explained here:
https://yandex.ru/dev/dialogs/alice/doc/session-persistence-docpage/

channels/yandex-alice/src/main/kotlin/com/justai/jaicf/channel/yandexalice/api/AliceBotRequest.kt
Pazus: The new field contains the **same** value `user_id` contained. The new place of the value is more semantically correct representation of what the value actually means. `application_id` is changed when a user re-logins.

PR #23, opened 2020-06-30
BotContextManager refactoring, context manager overriding support

morfeusys:
Here is an implementation of channel-related context manager functionality.
The usage as simple as

```
fun main() {
    embeddedServer(Netty, System.getenv("PORT")?.toInt() ?: 8080) {
        routing {
            httpBotRouting("/" to AliceChannel(skill, useDataStorage = true))
        }
    }.start(wait = true)
}
```

And then you can use the same transparent persistence in your scenarios:

```
context.session["first"] = Product(...)
context.client["last_reply"] = reply
```

and the same delegation in your data classes like

```
class MyContext(context: BotContext) {
  var name: String? by context.client
  var order: String? by context.session
}
```


channels/yandex-alice/README.md
```diff
@@ -364,7 +364,47 @@ state("repeat") {
...
+    activator.alice?.run {
+        val firstAmount = slots["first_amount"]
+        val secondAmount = slots["second_amount"]
+    }
+}
+```
+
+## Хранение состояния
+JAICF уже осуществляет автоматическое хранение состояния с помощью [различных БД](https://github.com/just-ai/jaicf-kotlin/wiki/Environments).
+Но вы также можете использовать механизм хранения состояния, [доступный в Яндекс Дислогах](https://yandex.ru/dev/dialogs/alice/doc/session-persistence-docpage/).
```
Denire, line 380: typo in "Дислогах"

PR #45, opened 2020-09-09
0.6.1 - fix CailaIntentActivator state recognition

Denire:
Fixes bug when CailaIntentActivator selects same intent if this intent confidence > 0

### DIFF
--- a/activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaIntentActivator.kt
+++ b/activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaIntentActivator.kt
@@ -55,9 +55,11 @@ class CailaIntentActivator(
             }
             // Sort all predicted intents by context relevance
             .sortedByDescending {
-                StatePath.parse(botContext.dialogContext.currentState)
-                    .stepUp().toString()
-                    .commonPrefixWith(it.first)
+                var currentState = botContext.dialogContext.currentState
+                if (currentState == it.first) {
+                    currentState = StatePath.parse(currentState).stepUp().toString()
+                }
+                currentState.commonPrefixWith(it.first)
             }
             .groupBy { it.first }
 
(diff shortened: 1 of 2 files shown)


nikvoloshin (approved):

activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaIntentActivator.kt
```diff
@@ -53,7 +54,11 @@ class CailaIntentActivator(
                 }
             }
             // Sort all predicted intents by context relevance
-            .sortedByDescending { botContext.dialogContext.currentState.commonPrefixWith(it.first) }
+            .sortedByDescending {
```
nikvoloshin, line 57: Now child states of the current state have the same "weights" as siblings. I suggest adding a comparison by context length for states with equal prefixes for increasing the "weight" of the child nodes.

nikvoloshin (changes requested):

activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaIntentActivator.kt
nikvoloshin: Actually, you can just replace comparison by prefix to comparison by context length (without `stepUp`), because nodes with lower common prefix length have lower context length.

nikvoloshin (approved):

activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaIntentActivator.kt
Denire: Unfortunately, I can't, because it will mess with global activators, where global activator if global activator is in different part of tree and has very long path.
I've changed a bit from PR and added fix for sibling selection. But this is still temporary and should be refactored with activators.

PR #47, opened 2020-09-12
Fix cleaning temp data after request processing finished

nikvoloshin:
This PR fixes a bug that `BotContext#temp` gets cleaned only when the new session starts. 
Now it is cleaned after a request processing finished.

Two things to notice:
- `temp` is not cleaned if slot filling in progress;
- `temp` is not cleaned if `BotRequestHook` interrupts request processing.

If it is not the desired logic, I'll fix it.

**EDIT:** Thanks, @Denire for pointing me out that `temp` data is actually gets cleaned implicitly as it's not saved after a request and so isn't loaded on the next request.
So the question is either we want to clean `temp` explicitly or remain implicit.

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
@@ -107,6 +107,7 @@ class BotEngine(
                 processStates(context)
                 saveContext(cm, botContext, request, reactions)
             }
+            botContext.cleanTempData()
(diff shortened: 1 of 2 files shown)


Denire:
I think it's better to make this logic explicit in BotEngine. Otherwise custom implementation of `BotContextManager` may provide inconsistency in `temp` data persistence.

Denire (approved):

Issue #38, opened 2020-09-01
Add ranking function to state selection


Denire:
Having multiple intents with different confidences, we have to select state for the most relevant intent, where relevance is defined by confidence + by context range (how close this new target state to current context). 
This issue suggests that we should create that ranking function and apply it during intent activation.

PR #51, opened 2020-09-21
[#38] Add ActivationStrategy and context ranking implementation

Denire:
This pull request adds ActivationStrategy interface and two implementations:
1. Default implementation - `ActivationByConfidence` - works like it was working before.
2. Ranking by context implementation - `ActivationByContextPenalty`.

Resolves https://github.com/just-ai/jaicf-kotlin/issues/38


core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByConfidence.kt
```diff
@@ -0,0 +1,39 @@
+package com.justai.jaicf.activator.strategy
+
+import com.justai.jaicf.context.BotContext
+import com.justai.jaicf.model.activation.Activation
+import com.justai.jaicf.model.activation.ActivationStrategy
+import com.justai.jaicf.model.state.StatePath
+
+/**
+ * Default implementation of [ActivationStrategy], used by [com.justai.jaicf.BotEngine]
```
nikvoloshin, line 9: **Should be fixed**
Let's remove **Default** word from this doc. 
The info about default implementation must be documented in `BotEngine` and not inside this concrete implementation.

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
+package com.justai.jaicf.activator.strategy
+
+import com.justai.jaicf.context.BotContext
+import com.justai.jaicf.helpers.logging.WithLogger
+import com.justai.jaicf.model.activation.Activation
+import com.justai.jaicf.model.activation.ActivationStrategy
+import com.justai.jaicf.model.state.StatePath
+
+/**
+ * Implementation of [ActivationStrategy] by analogy with JAICP DSL.
```
nikvoloshin, line 10: Why do we need the reference to JAICP here?
As for me, It's a redundant info that pollutes the doc

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * This activation strategy calculates penalty for each activation,
+ * applies penalties for every activation and selects single activation with highest adjusted confidence.
+ *
+ * Penalty increases with difference level between current state and activation target state.
```
nikvoloshin, line 15: **Should be fixed**
Describe more properly what is the difference between states

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * @property stepUpPenaltyBase penalty base for context stepUp.
+ * @see calculatePenalty method with formula
+ * */
+class ActivationByContextPenalty(
+    private val stepUpPenaltyBase: Double = 0.2
```
nikvoloshin, line 21: Could we name it just `penalty`?

core/src/main/kotlin/com/justai/jaicf/model/activation/ActivationStrategy.kt
```diff
@@ -0,0 +1,26 @@
...
+ * Abstraction for strategy, using which activation will be selected from list.
+ * */
+interface ActivationStrategy {
+
+    /**
+     * Selects [Activation] from list using implemented strategy.
+     *
+     * @param botContext a current user's [BotContext]
+     * @param activations list of all available activations
+     * @return the most relevant [Activation] in terms of certain implementation of [BaseActivator]
```
nikvoloshin, line 16: **Must be fixed**
As it's no longer related to an `Activator` implementation, let's rewrite this row.

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+) : ActivationStrategy, WithLogger {
+
+    internal data class ActivationWithScore(val activation: Activation, val adjustedConfidence: Double)
+
+    /**
+     * Selects activation by context similarity level and activation confidence.
```
nikvoloshin, line 27: Do we need to inherit detailed documentation? Why not just left with this row?

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByConfidence.kt
```diff
@@ -0,0 +1,39 @@
...
+ * of the current state, then from all siblings (including current state),
+ * then from all siblings of the parent (including the parent), and so on.
+ *
+ * @see com.justai.jaicf.BotEngine
+ * @see com.justai.jaicf.activator.Activator
+ * */
+object ActivationByConfidence : ActivationStrategy {
+
+     /*
+     * Selects activation by confidence using nearest states.
```
nikvoloshin, line 21: Do we need to inherit detailed documentation? Why not just left with this row?

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param botContext a current user's [BotContext]
+     * @param activations list of all available activations
+     * @return the most relevant [Activation] in terms of certain implementation of [BaseActivator]
+     *
+     * @see Activation
+     * @see ActivationStrategy
+     * @see com.justai.jaicf.BotEngine
+     */
+    override fun selectActivation(
```
nikvoloshin, line 37: **Should be fixed**
Two parameters can be on one line

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+    private fun logActivationResults(activationsByPenalty: List<ActivationWithScore>) = activationsByPenalty.take(5)
+        .joinToString("\n") { "to ${it.activation.state} with confidence ${it.adjustedConfidence}" }
+        .let { logger.debug("Possible transitions:\n$it") }
+
+    /**
+     * Applies context difference penalty for each activation.
+     *
+     * @param activations a list of possible activations
+     * @param currentState a state to find activations from
+     * @param activationStrategy configuration
```
nikvoloshin, line 57: **Must be fixed**
Redundant parameter

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+        val toState = targetState.components
+        val fromState = currentContext.components
+        var i = 0
+        while (i < fromState.size && i < toState.size) {
+            if (fromState[i] != toState[i]) {
+                break
+            }
+            i++
+        }
+        return fromState.size - i
```
nikvoloshin, line 94: Can be simplified to:
```
val commonPrefix = fromState.zip(toState).takeWhile { it.first == it.second }.count()
return fromState.size - commonPrefix
```

core/src/main/kotlin/com/justai/jaicf/activator/strategy/ActivationByContextPenalty.kt
```diff
@@ -0,0 +1,115 @@
...
+     * 1 - 0.8
+     * 2 - 0.7
+     * 3 - 0.63
+     * 4 - 0.58
+     *
+     * @param similarityLevel number of transitions between contexts
+     * @return 1 - x - x/2 - x/3 - x/4 - ... x/(l+1)
+     */
+    private fun calculatePenalty(similarityLevel: Int) =
+        (0 until similarityLevel).fold(1.0) { penalty, level ->
```
nikvoloshin, line 110: Can be simplified to `return 1 - stepUpPenaltyBase * (1..similarityLevel).map { 1.0 / it }.sum()`

core/src/test/kotlin/com/justai/jaicf/core/test/strategy/ActivationByContextPenaltyTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    private val currentNode = "/some/current/state"
+    private val childNode = "/some/current/state/child"
+    private val otherContextNode = "/another/target/state"
+    private val rootNode = "/"
+    private val siblingNode = "/some/current/sibling"
+
+    private val bc = BotContext("", DialogContext().apply { currentContext = currentNode })
+
+    @Test
+    fun shouldRankStrictActivations() {
```
nikvoloshin, line 25: Let's write more proper test names using backticks

nikvoloshin (changes requested):
@morfeusys, @Denire As far as I remember, we discussed some days ago that transition managing can be moved from `Activator` to `BotEngine` or somewhere. The point of making `Activator`s responsible for transition managing was the fact that different activator implementations can use different ranking mechanisms. Now, this responsibility is moved to `ActivationStrategy`, so why not to extract transition managing from `Activator`?

Denire:
> @morfeusys, @Denire As far as I remember, we discussed some days ago that transition managing can be moved from `Activator` to `BotEngine` or somewhere. The point of making `Activator`s responsible for transition managing was the fact that different activator implementations can use different ranking mechanisms. Now, this responsibility is moved to `ActivationStrategy`, so why not to extract transition managing from `Activator`?

This responsibility is not completely moved to `ActivationStrategy` (activationSelector). Some entity still has to generate transitions and create `Activation` objects from `ScenarioModel`
I'm not sure that we need to extract this new entity right now, as we don't have any other `ScenarioModel` even in ideas.

nikvoloshin:
> This responsibility is not completely moved to `ActivationStrategy` (activationSelector). Some entity still has to generate transitions and create `Activation` objects from `ScenarioModel`

What are entities that has to generate transitions on their own? After previous changes in activators, there is added a private method `BaseActivator.generateTransitions` that generates only **possible** transitions (and therefore it's private). No one custom activator should generate transitions on its own as it may break our internal logic.


> I'm not sure that we need to extract this new entity right now, as we don't have any other `ScenarioModel` even in ideas/issues.

It's not about another `ScenarioModel`, it's about Separation of Concerns and Single Responsibility. The current scenario model does not limit us in the building of proper architecture where each entity has only one concrete responsibility.

core/src/test/kotlin/com/justai/jaicf/core/test/strategy/ActivationByContextPenaltyTest.kt
nikvoloshin: It looks better if a function name in backticks starts with a capital letter

core/src/main/kotlin/com/justai/jaicf/activator/selection/ActivationByConfidence.kt
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * @param botContext a current user's [BotContext]
+     * @param activations list of all available activations
+     * @return the most relevant [Activation]
+     *
+     * @see Activation
+     * @see ActivationSelector
+     */
+    override fun selectActivation(botContext: BotContext, activations: List<Activation>): Activation {
+        val first = StatePath.parse(activations.first().state!!)
```
nikvoloshin, line 27: This implementation doesn't work too.
The previous implementation before changes relied on the fact, that the activations list is pre-sorted and the first element is always on the most relevant layer, that's why there was `takeWhile` that had taken only elements on the same level that the first is.
So the thing is you cannot rely on the `first` element.

Denire: Should be fixed now.

nikvoloshin: And this one doesn't work too, sorry :)
I guess something like that should work, but it needs testing
```
    override fun selectActivation(botContext: BotContext, activations: List<Activation>): Activation {
        val current = botContext.dialogContext.currentContext
        val (fromRoot, onPath) = activations.partition { it.state!!.commonPrefixWith(current) == "/" }
        val sorted = onPath.sortedByDescending { StatePath.parse(it.state!!).components.size } + fromRoot
        val first = StatePath.parse(sorted.first().state!!)
        return sorted.takeWhile {
            StatePath.parse(it.state!!).parent == first.parent
        }.maxBy { it.context.confidence }!!
    }
```
Briefly: we need to sort by context length first, but there can be global activators with a long context, so we split activations on ones that accessible directly from the root and ones that accessible from some state on the path from the current state to the root, and then sort adding all "from the root" states to the end.

core/src/main/kotlin/com/justai/jaicf/model/activation/ActivationSelector.kt
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Calculates context difference (number of transitions) between targetState and currentState
+     * e.g.: currentState = "/root/child1/child2/child3", targetState = "/root/child1"; diff level = 2
+     *
+     * @param targetState possible target state by activation
+     * @param currentContext current context
+     *
+     * @return number of transitions between current and target state
+     * */
+    protected fun calculateStatesDifference(targetState: StatePath, currentContext: StatePath): Int {
```
morfeusys, line 42: Why did you leave it protected? Looks like utility function that could be wrapped to companion.

core/src/main/kotlin/com/justai/jaicf/activator/Activator.kt
```diff
@@ -54,15 +55,18 @@ interface Activator {
...
      * @see BotContext
      * @see BotRequest
      * @see com.justai.jaicf.context.ActivatorContext
+     * @see ActivationSelector
      */
     fun activate(
         botContext: BotContext,
-        request: BotRequest
+        request: BotRequest,
+        activationSelector: ActivationSelector
```
morfeusys, line 69: Are you sure that selector should be passed to `Activator` for each `activate` invocation? Could it be any case where selector could be changed at runtime?

morfeusys (changes requested):

core/src/main/kotlin/com/justai/jaicf/activator/Activator.kt
Denire: It doesn't. But I don't like adding `ActivationSelector` as property to `Activator` interface. Not every `Activator` needs `activationSelector` as property. In fact, `Activator` implementation can operate without `ActivationSelector`. It's `Activator`'s implementation decision to use `ActivationSelector` or not. 

Adding `ActivationSelector` as property means that we should add it to `ActivatorFactory`, therefore making implementing optional logic even more strict.

So I decided to add activationSelector as argument. UPD: With default value.

core/src/main/kotlin/com/justai/jaicf/model/activation/ActivationSelector.kt
```diff
@@ -19,29 +20,44 @@ abstract class ActivationSelector {
...
-        activation.state?.let { targetState ->
-            calculateStatesDifference(StatePath.parse(targetState), currentState)
-        }
+/**
+ * Sorts list of activation by relevance to currentContext - first child nodes, then siblings, then parents, etc.
+ * */
+fun Collection<Activation>.sortedByContext(currentContext: String): List<Activation> {
+    val (fromRoot, onPath) = partition {
+        val isRoot = it.state!!.commonPrefixWith(currentContext) == "/"
+        val isIndirectChild = it.state.asStatePath().isIndirectChildOf(currentContext.asStatePath())
```
nikvoloshin, line 32: What if an indirect child comes not from the root, but from some of our ancestors? It should have a bigger priority, shouldn't it?
current = `/a/b`
indirect child = `/a/b/c/d`, fromState = `/a`

nikvoloshin (changes requested):

core/src/main/kotlin/com/justai/jaicf/model/activation/ActivationSelector.kt
Denire: At this point we don't know if indirect child was be activated by fromState or by global activator, and this issue cannot be solved with current architecture.

Looks like `ActivationSelector` should start to know about Transitions, should generate sorted transitions with or without penalty. Architectual rework will be done in further commits.

nikvoloshin: So, we are about to break our current logic (that is, by the way, considered as default in this PR) in order to add some other implementation. I don't think it should be done this way.

Denire: No, I meant that we shouldn't merge this PR until I've made changes, so we wont break current logic

nikvoloshin: Ah okay, I've misunderstood you, sorry.
And about your previous message. I don't think `ActivationSelector` should be aware of transitions or generate them on its own. We should provide all required data through an `Activation` object, precisely through some state-like object in it.

Denire: I think at some point some entity should take transition management from `Activator`, i thought about changing this right now, but it leads to way to many changes in architecture.

I think i'll implement your idea with providing `Activation` with necessary data. Thanks for idea!

Denire: @nikvoloshin I've added StatesTransition class holding information about transition (fromState, toState, distance). 

But there is much to improve in future. The problems I see now are:
1. checkStrictTransitions() in BotEngine class. It should not be there, also Strict Transition should be a separate entity.
2. Distance should be calculated when we create transition (generateTransitions).
3. I don't like that we have two classes (Transition and StatesTransition (or ActivationTransition?)) containing mostly the same data. Looks like we need only one of them.

But solving any of these problems requires a number of changes unrelated to issue. I'd like to solve them in separate PR in future.

nikvoloshin: @Denire Thanks for the changes! I like that now `Activation` has more info about states and transitions. 
One thing to discuss: it seems like we don't really need `StatesTransition` entity. We can pass the `Transition` object to the `Activation` and provide all utility functionality as extension functions because `StatesTransition` is a stateless and immutable object with `fromState` and `toState` as primary parameters.
So, please think in this direction, maybe it'd be better just to provide utility extension functions on `State` and `Transition` instead of creating utility-like entity

Denire: @nikvoloshin I thought about it and found two reasons not to use `Transition` object:

1. Transition has `ActivationRule` required property, and in case of Slot Filling or Strict Activations it will require unpleasant changes in BotEngine and other places, which don't care about `ActivationRule`. Their job is only to create and object and pass it further. 
Making `ActivationRule` optional also doesn't look good.

2. It may be better to use Transition for ScenarioBuilding and StatesTransition (ActivationTransition?) for processing activations. Therefore we have independent objects, and in case we need to modify one of them, we won't be affecting many classes, which use only ActivationTransition logic. 

upd: I'll rename StatesTransition to ActivationTransition.

Denire:
@nikvoloshin since last discussion i've created some changes. The most interesting part here is: 
```kotlin
    private fun generateTransitions(botContext: BotContext): Map<StatePath, List<Transition>> {
        val currentState = StatePath.parse(botContext.dialogContext.currentContext).resolve(".")
        return generateSequence(currentState) { if (it.toString() == "/") null else it.stepUp() }.map { statePath ->
            val transitions = transitions[statePath.toString()] ?: emptyList()
            statePath to transitions
        }.toMap()
    }
```

As we discussed, we'll make architectural changes to Activation, mb add ActivationManager, or some entity to manage activation transitions. But now I decided to make least effort to get the result. 

Well, at least when we will make refactoring, we'll have some integration tests now :)

morfeusys (changes requested):

Denire:
Closed as it was solved in separate pr

PR #54, opened 2020-09-23
Implement anyIntent activation rule.

### LINKED ISSUE #52: Add match event
NLU Providers can contain default answer for intent, which can be extracted from ActivatonContext in scenario and responded. Example usage:
```kotlin
state("handler") {
    activators {
        intent("HowAreYou")
    }
    action {
        activator.caila?.topIntent?.answer?.let {
            reactions.say(it)
        }
    }
}
```

It looks ok when there's not so many intents. But when intents number increases, usability decreases. User should list all project intents in single state and he cannot separate dialogue logic from scenario.
```kotlin
state("handler") {
    activators {
        intent("HowAreYou")
        intent("WhatCanYouDo")
        intent("Family")
        intent("WhatAreYouDoing")
        ...
    }
    action {
        activator.caila?.topIntent?.answer?.let {
            reactions.say(it)
        }
    }
}
```

This issue suggest adding something like:
```kotlin
state("handler") {
    activators {
        event(Activations.AnyActivation)
    }
    action {
        activator.caila?.topIntent?.answer?.let {
            reactions.say(it)
        }
    }
}
```

nikvoloshin:
This PR adds support to `anyIntent` activation rule.
A state with such a rule in its `activators {}` block will be activated if any intent activator recognized any intent.

Resolves #52

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
@@ -1,7 +1,8 @@
 package com.justai.jaicf.builder
 
 import com.justai.jaicf.activator.catchall.CatchAllActivationRule
-import com.justai.jaicf.activator.event.EventActivationRule
+import com.justai.jaicf.activator.event.AnyEventActivationRule
+import com.justai.jaicf.activator.event.EventByNameActivationRule
 import com.justai.jaicf.activator.intent.AnyIntentActivationRule
 import com.justai.jaicf.activator.intent.IntentByNameActivationRule
 import com.justai.jaicf.activator.regex.RegexActivationRule
@@ -111,7 +112,7 @@ abstract class ScenarioBuilder(
         sb.body()
 
         if (model.states[sb.path.toString()] != null) {
-            throw IllegalStateException("Dublicated declaration of state with path: ${sb.path}")
+            throw IllegalStateException("Duplicated declaration of state with path: ${sb.path}")
         }
 
         model.states[sb.path.toString()] = sb.build()
@@ -119,37 +120,6 @@ abstract class ScenarioBuilder(
         statesStack.removeLast()
         currentState = statesStack.last
     }
-
-    /**
-     * Appends an any-intent state to the scenario.
-     * This state will be activated for every request that has any intent in its input, recognized by
-     * any registered IntentActivator.
-     *
-     * The current dialogue's context won't be changed.
-     * This builder requires an IntentActivator to be added to the activators list of your BotEngine instance.
-     *
-     * ```
-     * anyIntent {
-     *   activator.caila?.topIntent?.answer?.let {
-     *       reactions.say(it)
-     *   }
-     * }
-     * ```
-     *
-     * @param state an optional state name ("anyIntent" by default)
-     * @param action an action block that will be executed
-     */
-    fun anyIntent(
-        state: String = "anyIntent",
-        action: ActionContext.() -> Unit
-    ) = state(
-        name = state,
-        noContext = true,
-        body = {
-            activators { anyIntent() }
-            action(action)
-        }
-    )
 
     /**
      * Appends a fallback state to the scenario.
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
```diff
@@ -276,7 +279,21 @@ abstract class ScenarioBuilder(
...
+        ))
+
+        /**
+         * Appends any-intent activator to this state. Means that any intent can activate this state.
+         * Requires a [com.justai.jaicf.activator.intent.IntentActivator] in the activators' list of your [com.justai.jaicf.api.BotApi] instance.
+         *
+         * @see com.justai.jaicf.activator.intent.IntentActivator
+         * @see com.justai.jaicf.api.BotApi
+         */
+        fun anyIntent() = add(
```
Denire, line 292: I think we need to make anyIntent internal and provide like-fallback builder:
```kotlin
    fun anyIntent(
        state: String = "anyIntent",
        action: ActionContext.() -> Unit
    ) = state(
        name = state,
        noContext = true,
        body = {
            activators { anyIntent() }
            action(action)
        }
    )
```

to use something like

```kotlin
        state("/someIntent") {
           ...
        }

        anyIntent {
            activator.caila?.topIntent?.answer?.let { 
                reactions.say(it)
            }
        }
```

nikvoloshin: Great idea!
Can you propose a name for this "state-like" builder function? I don't want both methods (builder and rule) to have the same name.

nikvoloshin: We can use `intent()` without argument as _any-intent_ rule, and `anyIntent {}` as state builder. What do you think?

Denire: Maybe we just make `anyIntent()` rule internal, therefore only builder will be available in scenario?
I see no point making rule public, this will only give user way to write bad code, giving not so much of benefits.

Denire:
```kotlin
intent() {
    .. anyIntent logics
}
```
looks a bit counter-intuitive, and also I think we'll be adding more builders to scenario. For example, a builder for state with single intent activator (it looks to be a fairly common use case). We'd better to keep this method signature reserved.
``` kotlin
fun intent(
        intent: String,
        action: ActionContext.() -> Unit
    ) = state(
        name = intent,
        noContext = true,
        body = {
            activators { intent(intent) }
            action(action)
        }
    )
```
to optimise scenario building with states with only one intent activator (i think that's most of the cases). E.g.:
```kotlin
        intent("Hello) {
            reactions.say("Hi there!")
        }
```

Denire (approved):
Looks good. Thank you!

core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
```diff
@@ -117,6 +120,37 @@ abstract class ScenarioBuilder(
...
+     *   activator.caila?.topIntent?.answer?.let {
+     *       reactions.say(it)
+     *   }
+     * }
+     * ```
+     *
+     * @param state an optional state name ("anyIntent" by default)
+     * @param action an action block that will be executed
+     */
+    fun anyIntent(
```
morfeusys, line 142: @nikvoloshin Good enough for the very first implementation. We could think about more general `unhandled` activation, that means "Activate on any input including _intent_, _event_ or _anything else_ that cannot be handled by any other state". How do you think?

core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
```diff
@@ -276,7 +310,20 @@ abstract class ScenarioBuilder(
...
+        ))
+
+        /**
+         * Appends any-intent activator to this state. Means that any intent can activate this state.
+         * Requires a [com.justai.jaicf.activator.intent.IntentActivator] in the activators' list of your [com.justai.jaicf.api.BotApi] instance.
+         *
+         * @see com.justai.jaicf.activator.intent.IntentActivator
+         * @see com.justai.jaicf.api.BotApi
+         */
+        internal fun anyIntent() = add(Transition(
```
morfeusys, line 323: Are you sure it **should** be internal? Let's think about how the user could use this activation in other cases if they would like to activate some state that is already bound to any other activators.

morfeusys (commented):
@nikvoloshin @Denire is it a time to get rid of `skippedActivators` once we have now a mechanism to handle a single use-case that requires this field?

core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
nikvoloshin: As far as I understand, it's exactly what `catchAll` activator means, isn't is?
If so, does that mean that we want to unify `catchAll`, `anyIntent` or anything else with similar logic to one `unhandled`-like rule?

nikvoloshin:
- Add `anyEvent` activation rule that works as `anyIntent`
- Make `anyIntent`, `anyEvent` rules public
- Remove state-builder `anyIntent` (I think, activation rules are enough for now)

morfeusys (approved):

Issue #49, opened 2020-09-14
ActionErrorHook cannot prevent exception from throwing


Denire:
`ActionErrorHook` is a hook invoked on action error. But use case for this hook looks a bit unclear.

Having error, user might want to handle this error, but hook does not allow error handling. It only allows to react in some way to error, while further processing in `BotEngine` will be aborted.

@morfeusys maybe we should create `ActionErrorHandler` instead of `ActionErrorHook`?

morfeusys:
@Denire maybe it's better to fix an implementation of BotEngine that should to work with this type of hook the same way as with others?

PR #58, opened 2020-09-30
[#49] remove exception throwing in BotEngine, add tests for ActionErrorHook

just-authorization:
On action exception we don't need to throw error outside or BotEngine. There was issues when exceptions in `action` block crashed `TelegramChannel` as there was no handlers inside.
This PR should fix it.


just-authorization:
Oops. Wrong account.

PR #59, opened 2020-09-30
[#49] remove exception throwing in BotEngine, add tests for ActionErrorHook

Denire:
On action exception we don't need to throw error outside or `BotEngine`. There was issues when exceptions in `action` block crashed `TelegramChannel` as there was no handlers inside.
This PR should fix it.

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
@@ -234,6 +234,7 @@ class BotEngine(
                                 state.action.execute(this)
                                 withHook(AfterActionHook(botContext, request, reactions, activator, state))
                             } catch (e: Exception) {
+                                logger.warn("Action exception on state ${dc.currentState}", e)
                                 hooks.triggerHook(ActionErrorHook(botContext, request, reactions, activator, state, e))
                             }
                         }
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
```diff
@@ -234,8 +234,8 @@ class BotEngine(
                                 state.action.execute(this)
                                 withHook(AfterActionHook(botContext, request, reactions, activator, state))
                             } catch (e: Exception) {
+                                logger.warn("Action exception on state ${dc.currentState}", e)
```
nikvoloshin, line 237: Let's use `error` log level for exception logging

Denire: I thought about it. In many cases `error` level is reserved for something critical. I don't think here we should set `error`, just like when we've not found a state, and user request cannot be processed, we use `logger.warn()`.

nikvoloshin: I think, that skipping user's action is critical enough as it can cause problems with not filled properly context or not executed `go` that can make user stuck in some state, and so on. As soon as not found state will cause only skipping single user's request and will not affect scenario logic.
So I'd better log it under `error` level

nikvoloshin (approved):

morfeusys (approved):

PR #66, opened 2020-10-05
Disable step up and implement ranking

### LINKED ISSUE #38: Add ranking function to state selection

nikvoloshin:
- Now only the following transitions are available: 
  * transitions from the current state to its children
  * transitions to the current state itself
  * transitions from the root, if the current state is not modal

- Implemented activation selection with two default implementations:
  * by the context and then by the confidence
  * by confidence weighted with the context

Closes #38

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/model/activation/selection/ActivationSelector.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/model/activation/selection/ActivationSelector.kt
@@ -1,0 +1,38 @@
+package com.justai.jaicf.model.activation.selection
+
+import com.justai.jaicf.context.ActivatorContext
+import com.justai.jaicf.context.BotContext
+import com.justai.jaicf.model.activation.Activation
+import com.justai.jaicf.model.transition.Transition
+
+/**
+ * A selector, that is used for selecting the most relevant [Activation] from all possible.
+ *
+ * @see [com.justai.jaicf.activator.Activator]
+ * @see [ContextFirstActivationSelector]
+ * @see [ContextRankingActivationSelector]
+ */
+interface ActivationSelector {
+
+    /**
+     * Receives a list of all possible transitions and selects the most relevant one.
+     *
+     * @param botContext a current [BotContext].
+     * @param activations a list of pairs of [Transition] and [ActivatorContext] to choose from.
+     *
+     * @return [Activation] that is built from the most relevant transition and context,
+     * or `null` if no [Activation] should be selected.
+     */
+    fun selectActivation(
+        botContext: BotContext,
+        activations: List<Pair<Transition, ActivatorContext>>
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/com/justai/jaicf/model/activation/selection/ContextRankingActivationSelector.kt
```diff
@@ -0,0 +1,49 @@
...
+     * One can override this method and implement its own scoring mechanism.
+     */
+    protected open fun calculateScore(
+        botContext: BotContext,
+        transition: Transition,
+        context: ActivatorContext
+    ): Float {
+        val current = botContext.dialogContext.currentContext
+
+        val factor = when {
```
morfeusys, line 40: Why these magic numbers?

core/src/main/kotlin/com/justai/jaicf/model/activation/selection/ActivationSelector.kt
```diff
@@ -0,0 +1,38 @@
+package com.justai.jaicf.model.activation.selection
```
morfeusys, line 1: Are you sure this interface and its implementations should be placed to `model`?

core/src/main/kotlin/com/justai/jaicf/model/activation/selection/ContextRankingActivationSelector.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Selects the most relevant activation, based on confidence weighting.
+ *
+ * Default implementation multiplies activation confidence by a factor that is different
+ * for different layers of scenario graph. More specific transition means greater factor.
+ * Specificity is ordered as follows:
+ * 1. Children of the current state;
+ * 2. The current state itself;
+ * 3. States available from the root of scenario.
+ */
+open class ContextRankingActivationSelector : ActivationSelector {
```
morfeusys, line 18: Why not to use this implementation as default selector instead of `ContextFirstActivationSelector`? Looks like it covers the context-first selector cases.

morfeusys (changes requested):

core/src/main/kotlin/com/justai/jaicf/model/activation/selection/ContextRankingActivationSelector.kt
nikvoloshin: Just by _trial-and-error_. We have only three sets of states, so we can define constants for them instead of calculating some penalty at runtime

morfeusys (approved):

PR #67, opened 2020-10-05
Fix the bug with the RegexActivator when unmatched group cause exception

### LINKED ISSUE #62: RegexActivator: exception is thrown if group hadn't matched input
As far as I understand, `Matcher#group(id)` will return null, if a group hadn't matched any input.
And as we put groups in non-nullable `RegexActicvatorContext#groups` list, `IllegalStateException` occurres.

nikvoloshin:
This changes make `RegexActivatorContext#groups` and `RegexActivatorContext#namedGroups` hold nullable values, as `Matcher.group(id)` will return null on unmatched group.

Closes #62


nikvoloshin:
Wrong history. Reopened as #68

PR #85, opened 2020-10-15
Bundle BaseEventActivator and BaseIntentActivator

nikvoloshin:
`BaseEventActivator` and `BaseIntentActivator` go to the end of the activators list if were not specified directly by the user.
`CatchAllActivator` always goes to the end, even if was specified in another place.

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
@@ -5,6 +5,8 @@ import com.justai.jaicf.activator.Activator
 import com.justai.jaicf.activator.ActivatorFactory
 import com.justai.jaicf.activator.selection.ActivationSelector
 import com.justai.jaicf.activator.catchall.CatchAllActivator
+import com.justai.jaicf.activator.event.BaseEventActivator
+import com.justai.jaicf.activator.intent.BaseIntentActivator
 import com.justai.jaicf.api.BotApi
 import com.justai.jaicf.api.BotRequest
 import com.justai.jaicf.api.hasQuery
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
```diff
@@ -63,12 +65,19 @@ class BotEngine(
...
-        } else {
-            it
+    private val activators = activators.map { it.create(model) }.addBuiltinActivators()
+
+    private fun List<Activator>.addBuiltinActivators(): List<Activator> {
+        fun MutableList<Activator>.removeIfPresent(a: Activator) = removeIf { it.name == a.name }
+        fun MutableList<Activator>.addIfAbsent(a: Activator) = find { it.name == a.name } ?: add(a)
+
+        val catchAll = CatchAllActivator.create(model)
+        val builtinActivators = listOf(BaseEventActivator, BaseIntentActivator).map { it.create(model) }
```
morfeusys, line 75: Why not to add `CatchAllActivator` here and try to remove if present all of these activators?

morfeusys (changes requested):

core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
nikvoloshin: Basically, I don't want to force our order if the user provides activators by himself in a certain order, and I'd better do `addIfAbsent` for all of these activators without rearranging them.
But what if the user puts `CatchAllActivator` in the middle of the list? Where then `BaseEvent` and `BaseIntent` should go, before or after `CatchAll`?
So, I decided to consider `CatchAllActivator` as a special one that must always be present and must always be at the end of the list. At the same time, other activators will go to the end (before `CatchAll`) only if the user hasn't specified the order directly.

morfeusys (approved):

Denire (approved):

PR #87, opened 2020-10-21
Gradle plugin for JAICP Cloud deployment

nikvoloshin:
The plugin works as follows:
- Applies `ShadowJar` plugin
- Defines `jaicpBuild` task
- `jaicpBuild` task, when called, reconfigures the `shadowJar` task
- Then reconfigured `shadowJar` task runs right after `jaicpBuild` task

If the `jaicpDeploy` task was not called, `shadowJar` will not be reconfigured, so a user can use his own `shadowJar` configuration for local development.

It is possible to provide custom `jarDestinationDir` and `jarFileName` during deployment as follows:
```
./gradew jaicpBuild -Pcom.justai.jaicf.jaicp.deploy.jarDestinationDir=some/dir \
-Pcom.justai.jaicf.jaicp.deploy.jarFileName=some-file-name.jar
```


gradle-plugins/jaicp-deploy-plugin/src/main/kotlin/com/justai/jaicf/plugins/jaicpdeploy/GradleUtils.kt
```diff
@@ -0,0 +1,20 @@
+package com.justai.jaicf.plugins.jaicpdeploy
```
Denire, line 1: It's quite possible that we'll be adding more plugins for JAICP integration, so I propose using package like
`com.justai.jaicf.plugins.jaicp.build`
Plus, I think this plugin should be `build`. At some point in future could like create `deploy` plugin, which will deploy project from IDEA to JAICP  Cloud directly.

gradle-plugins/jaicp-deploy-plugin/src/main/kotlin/com/justai/jaicf/plugins/jaicpdeploy/JaicpDeployPlugin.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @TaskAction
+    fun action() {
+        project.tasks.withType(ShadowJar::class.java).named(SHADOW_JAR_TASK_NAME) { jar ->
+            jar.destinationDirectory.set(project.file(jarDestinationDir))
+            jar.archiveFileName.set(jarFileName)
+
+            val mainClass = mainClassName ?: jar.manifest.attributes["Main-Class"]
+            mainClass?.let { jar.manifest.attributes["Main-Class"] = it }
+                ?: project.logger.warn("Main class name was not provided")
```
Denire, line 43: Shouldn't we fail and throw exception when main class name was not provided?

Denire (changes requested):

gradle-plugins/jaicp-deploy-plugin/src/main/kotlin/com/justai/jaicf/plugins/jaicpdeploy/JaicpDeployPlugin.kt
nikvoloshin: I think we shouldn't, because the Gradle jar task is not the only way to set the main class. A user can have some `dependsOn`, or `finalizedBy` tasks, that set proper main class.

Denire (approved):

morfeusys (approved):

PR #88, opened 2020-10-22
jaicp slack integration

Denire:
This PR fixes JAICP integration with Slack

### DIFF
--- a/channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
+++ b/channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
@@ -6,6 +6,7 @@ import com.justai.jaicf.channel.http.HttpBotResponse
 import com.justai.jaicf.channel.jaicp.JaicpCompatibleAsyncBotChannel
 import com.justai.jaicf.channel.jaicp.JaicpCompatibleAsyncChannelFactory
 import com.justai.jaicf.context.RequestContext
+import com.justai.jaicf.helpers.http.toUrl
 import com.justai.jaicf.helpers.kotlin.PropertyWithBackingField
 import com.slack.api.Slack
 import com.slack.api.SlackConfig
@@ -39,7 +40,7 @@ class SlackChannel private constructor(
 
     private constructor(botApi: BotApi, urlPrefix: String): this(botApi) {
         val config = SlackConfig().apply {
-            methodsEndpointUrlPrefix = urlPrefix
+            methodsEndpointUrlPrefix = "$urlPrefix/".toUrl()
             methodsConfig = MethodsConfig().apply {
                 isStatsEnabled = false
             }
@@ -112,7 +113,7 @@ class SlackChannel private constructor(
 
     companion object : JaicpCompatibleAsyncChannelFactory {
         override val channelType = "slack"
-        override fun create(botApi: BotApi, apiUrl: String) = SlackChannel(botApi, "$apiUrl/")
+        override fun create(botApi: BotApi, apiUrl: String) = SlackChannel(botApi, apiUrl)
     }
 }
 
--- a/channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
+++ b/channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
@@ -29,16 +29,16 @@ class SlackChannel private constructor(
     private lateinit var app: App
     private lateinit var parser: SlackRequestParser
 
-    constructor(botApi: BotApi, config: SlackChannelConfig): this(botApi) {
-        app = App(AppConfig.builder()
+    constructor(botApi: BotApi, config: SlackChannelConfig) : this(botApi) {
+        val appConfig = AppConfig.builder()
             .singleTeamBotToken(config.botToken)
             .signingSecret(config.signingSecret)
             .build()
-        )
+        app = App(appConfig, config.middleware)
         start()
     }
 
-    private constructor(botApi: BotApi, urlPrefix: String): this(botApi) {
+    private constructor(botApi: BotApi, urlPrefix: String) : this(botApi) {
         val config = SlackConfig().apply {
             methodsEndpointUrlPrefix = "$urlPrefix/".toUrl()
             methodsConfig = MethodsConfig().apply {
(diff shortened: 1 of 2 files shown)


channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
```diff
@@ -99,7 +112,7 @@ class SlackChannel private constructor(
 
     companion object : JaicpCompatibleAsyncChannelFactory {
         override val channelType = "slack"
-        override fun create(botApi: BotApi, apiUrl: String) = SlackChannel(botApi, apiUrl)
+        override fun create(botApi: BotApi, apiUrl: String) = SlackChannel(botApi, "$apiUrl/")
```
morfeusys, line 115: I'd append slash inside constructor, incapsulating this logic and keeping factory clean

nikvoloshin, line 115: I think we shouldn't just put a slash at the end and make a user think more about how URL should look like.
Let's just do something like `apiUrl.takeIf { ie.endsWith("/") } ?: apiUrl + "/"`

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
```diff
@@ -38,10 +40,21 @@ class SlackChannel private constructor(
     private constructor(botApi: BotApi, urlPrefix: String): this(botApi) {
         val config = SlackConfig().apply {
             methodsEndpointUrlPrefix = urlPrefix
+            methodsConfig = MethodsConfig().apply {
+                isStatsEnabled = false
```
morfeusys, line 44: How this option affects requests processing?

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
```diff
@@ -38,10 +40,21 @@ class SlackChannel private constructor(
...
+            }
         }
 
         val slack = Slack.getInstance(config)
-        app = App(AppConfig.builder().slack(slack).build())
+        app = App(AppConfig.builder()
+            .slack(slack)
+            .alwaysRequestUserTokenNeeded(false)
+            .singleTeamBotToken("empty")
```
morfeusys, line 52: Why not to leave it empty?

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
```diff
@@ -38,10 +40,21 @@ class SlackChannel private constructor(
...
         val config = SlackConfig().apply {
             methodsEndpointUrlPrefix = urlPrefix
+            methodsConfig = MethodsConfig().apply {
+                isStatsEnabled = false
+            }
         }
 
         val slack = Slack.getInstance(config)
-        app = App(AppConfig.builder().slack(slack).build())
+        app = App(AppConfig.builder()
```
nikvoloshin, line 49: Should we do this configuration by ourselves? Is there a way to let a user configure the channel?

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
```diff
@@ -38,10 +40,21 @@ class SlackChannel private constructor(
     private constructor(botApi: BotApi, urlPrefix: String): this(botApi) {
```
morfeusys, line 40: @Denire What do you think about providing the constructor that accepts a complete Slack's `AppConfig`?

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
```diff
@@ -38,10 +40,21 @@ class SlackChannel private constructor(
...
 
         val slack = Slack.getInstance(config)
-        app = App(AppConfig.builder().slack(slack).build())
+        app = App(AppConfig.builder()
+            .slack(slack)
+            .alwaysRequestUserTokenNeeded(false)
+            .singleTeamBotToken("empty")
+            .signingSecret("empty")
+            .build(),
+            listOf(IgnoringSelfEvents(config))
```
morfeusys, line 55: Why not to add this list to the first constructor?

morfeusys (changes requested):

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
morfeusys: @nikvoloshin we have `String.toUrl()` utility function that fixes all double slashes in the URL

nikvoloshin: @morfeusys Thanks for pointing that out! 
Then it should be called inside constructor: `methodsEndpointUrlPrefix = urlPrefix.toUrl()`.

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
Denire: This constructor is invoked from JAICP channel factory, and it is configured to process messages without OAuth, signature and etc. This is our internal logic concerning integration with JAICP, so it should not be configured externally, as I think.

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
Denire: This disables OAuth authorization on every request, as we don't have OAauth configured in JAICP.

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
Denire: There are checks for empty strings in internal Slack filters. Leaving it empty or null will cause errors.

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
Denire: Because this filter is provided by default in AppConfig.

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
Denire: Not sure. At least not now. 

I don't see any reasons to recongifure AppConfig as it works OK now with defaults, and as we use only OAuth authorization in our channel.

channels/slack/src/main/kotlin/com/justai/jaicf/channel/slack/SlackChannel.kt
nikvoloshin: Okay, then maybe we should give a user a way to configure the channel inside another constructor? 
`com.justai.jaicf.channel.slack.SlackChannelConfig` seems like a very weak configurability

Denire: I agree. Middleware was added to SlackChannelConfig.

nikvoloshin (approved):

morfeusys (approved):

PR #95, opened 2020-10-26
add Reactions.sayRandom for List<T>

discodasha:
sometimes there's need to use sayRandom reaction for list of strings

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
@@ -33,8 +33,7 @@ open class ActionContext(
      * @param elements a list of elements to select random from
      * @return a random element
      */
-    private fun <T> random(elements: List<T>) = elements[smartRandom(elements.size, this) % elements.size]
-
+    fun <T> random(elements: List<T>) = elements[smartRandom(elements.size, this) % elements.size]
 
     /**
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
```diff
@@ -30,10 +30,19 @@ open class ActionContext(
     /**
      * Returns random element.
      * Works as a smart random preserving previous calls results to guarantee that there are no two equal results in two consecutive calls.
+     * @param elements a list of elements to select random from
+     * @return a random element
+     */
+    private fun <T> random(elements: List<T>) = elements[smartRandom(elements.size, this) % elements.size]
```
Denire, line 36: This function can be useful, i suggest removing private modifier.

core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
```diff
@@ -43,6 +52,7 @@ open class ActionContext(
      */
     open fun random(max: Int) = smartRandom(max, this)
 
+
```
Denire, line 55: It's better to stick to one line spacing between functions.

Denire (changes requested):

Denire (approved):

PR #101, opened 2020-10-29
query params provider and slot filling

morfeusys:
This PR appends two more features to Dialogflow:

* slot filling support
* query parameters provider

### DIFF
--- a/activators/dialogflow/src/main/kotlin/com/justai/jaicf/activator/dialogflow/DialogflowIntentActivator.kt
+++ b/activators/dialogflow/src/main/kotlin/com/justai/jaicf/activator/dialogflow/DialogflowIntentActivator.kt
@@ -1,18 +1,21 @@
 package com.justai.jaicf.activator.dialogflow
 
-import com.google.cloud.dialogflow.v2.QueryParameters
 import com.justai.jaicf.activator.Activator
 import com.justai.jaicf.activator.ActivatorFactory
 import com.justai.jaicf.activator.intent.BaseIntentActivator
 import com.justai.jaicf.api.BotRequest
 import com.justai.jaicf.api.hasEvent
 import com.justai.jaicf.api.hasQuery
+import com.justai.jaicf.context.ActivatorContext
 import com.justai.jaicf.context.BotContext
 import com.justai.jaicf.model.scenario.ScenarioModel
+import com.justai.jaicf.reactions.Reactions
+import com.justai.jaicf.slotfilling.*
 
 class DialogflowIntentActivator(
     model: ScenarioModel,
-    private val connector: DialogflowConnector
+    private val connector: DialogflowConnector,
+    private val queryParametersProvider: QueryParametersProvider = QueryParametersProvider.default
 ): BaseIntentActivator(model) {
 
     override val name = "dialogflowIntentActivator"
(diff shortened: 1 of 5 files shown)


Denire (approved):

activators/dialogflow/src/main/kotlin/com/justai/jaicf/activator/dialogflow/DialogflowIntentActivator.kt
```diff
@@ -44,9 +39,47 @@ class DialogflowIntentActivator(
...
+        var context = activatorContext?.dialogflow?.apply {
+            if (queryResult.allRequiredParamsPresent) {
+                return SlotFillingSkipped
+            }
+        }
+        if (context == null) {
+            context = recogniseIntent(botContext, request).firstOrNull()
+        }
+        return when {
+            context == null -> SlotFillingInterrupted
```
nikvoloshin, line 58: It seems like if an intent has required parameters, then on the first "interruption" by user, dialogflow will interrupt slotfilling, but will return the same intent in the response. So, in the current implementation, we will return `SlotFillingInProgress` in this case, as I understand

User: Hello
Dialogflow: "HelloIntent", action = "user_defined_slotfilling_action_name"
Bot: Hi, what is your name? (required) 
User: Cancel
Dialogflow: "HelloIntent", action = null
User: Cancel
Dialogflow: "Fallback"

===================================

Also, we should check `queryResult.isFallback`, because even in the case of fallback Dialogflow will return intent, and context will not be null. I guess, this check should be done in `recognizeIntent`

activators/dialogflow/src/main/kotlin/com/justai/jaicf/activator/dialogflow/DialogflowIntentActivator.kt
```diff
@@ -44,9 +39,47 @@ class DialogflowIntentActivator(
...
+            context.queryResult.allRequiredParamsPresent -> SlotFillingFinished(context)
```
nikvoloshin, line 59: Shouldn't we respond with Dialogflow prompt message? The same for `SlotFillingSkipped` case.

-> Hello
<- Hi, what is your name?
-> John
<- Nice to meet you, John! (will not be displayed)

nikvoloshin (changes requested):
Also, an important thing not about just slot-filling, but about stateful NLU services at all, one of the reasons why I'm stuck with Lex.  We should somehow manage the outside context. 
The first case (currently not supported), what if we've done `recognizeIntent`, but then decided not to stay with this activator? We should then clean the outside context for the next request.
The second case (more realistic), what if during the slot-filling process user has changed intent? Then we should interrupt slot-filling AND clear the context for the next request.
In other words, every time we are done (either with success or failure) with the current intent in a bot, we should inform NLU.

activators/dialogflow/src/main/kotlin/com/justai/jaicf/activator/dialogflow/DialogflowIntentActivator.kt
morfeusys: No, in this case we return `SlotFillingFinished`. You can just check the current example using "wake up" input to trigger the slot filling and then send "cancel".

nikvoloshin (approved):

PR #111, opened 2020-11-11
Fix Telegram buttons

### LINKED ISSUE #110: Fix Telegram buttons
Now it's impossible to use `reactions.buttons()` with Telegram. Because Telegram Telegram allows only to send keyboard attached to some message.

The idea is to use Telegram's `editMessageReplyMarkup` appending inline buttons to the previous message sent to the channel.

morfeusys:
fixes #110

### DIFF
--- a/channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramBotRequest.kt
+++ b/channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramBotRequest.kt
@@ -1,8 +1,7 @@
 package com.justai.jaicf.channel.telegram
 
-import com.github.kotlintelegrambot.entities.Contact
-import com.github.kotlintelegrambot.entities.Location
-import com.github.kotlintelegrambot.entities.Message
+import com.github.kotlintelegrambot.entities.*
+import com.github.kotlintelegrambot.entities.stickers.Sticker
 import com.justai.jaicf.api.BotRequest
 import com.justai.jaicf.api.EventBotRequest
 import com.justai.jaicf.api.QueryBotRequest
@@ -15,6 +14,9 @@ val TelegramBotRequest.location
 
 val TelegramBotRequest.contact
     get() = this as? TelegramContactRequest
+
+internal val Message.clientId
+    get() = from?.id?.toString() ?: chat.id.toString()
 
 interface TelegramBotRequest: BotRequest {
     val message: Message
@@ -54,5 +56,74 @@ data class TelegramContactRequest(
     input = TelegramEvent.CONTACT
 )
 
-internal val Message.clientId
-    get() = from?.id?.toString() ?: chat.id.toString()
+data class TelegramAudioRequest(
+    override val message: Message,
+    val audio: Audio
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.AUDIO
+)
+
+data class TelegramDocumentRequest(
+    override val message: Message,
+    val document: Document
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.DOCUMENT
+)
+
+data class TelegramAnimationRequest(
+    override val message: Message,
+    val animation: Animation
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.ANIMATION
+)
+
+data class TelegramGameRequest(
+    override val message: Message,
+    val game: Game
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.GAME
+)
+
+data class TelegramPhotosRequest(
+    override val message: Message,
+    val photos: List<PhotoSize>
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.PHOTOS
+)
+
+data class TelegramStickerRequest(
+    override val message: Message,
+    val sticker: Sticker
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.STICKER
+)
+
+data class TelegramVideoRequest(
+    override val message: Message,
+    val video: Video
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.VIDEO
+)
+
+data class TelegramVideoNoteRequest(
+    override val message: Message,
+    val videoNote: VideoNote
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.VIDEO_NOTE
+)
+
+data class TelegramVoiceRequest(
+    override val message: Message,
+    val voice: Voice
+): TelegramBotRequest, EventBotRequest(
+    clientId = message.clientId,
+    input = TelegramEvent.VOICE
+)
--- a/channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramBotRequest.kt
+++ b/channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramBotRequest.kt
@@ -16,7 +16,7 @@ val TelegramBotRequest.contact
     get() = this as? TelegramContactRequest
 
 internal val Message.clientId
-    get() = from?.id?.toString() ?: chat.id.toString()
+    get() = chat.id.toString()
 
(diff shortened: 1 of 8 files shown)


examples/hello-world/src/main/kotlin/com/justai/jaicf/examples/helloworld/channel/Telegram.kt
```diff
@@ -4,5 +4,5 @@ import com.justai.jaicf.channel.telegram.TelegramChannel
 import com.justai.jaicf.examples.helloworld.helloWorldBot
 
 fun main() {
-    TelegramChannel(helloWorldBot, "580468601:AAHaMg4gOsN2A_zvIO6-PouVk3GcZ_WVrdI").run()
+    TelegramChannel(helloWorldBot, "580468601:AAFZaIXLiKbD7XRsGVHX74N-5nlXbttvEzc").run()
```
Denire, line 7: Is it a real token?
Maybe we should host a bot somewhere (heroku/jaicp cloud) with this token to be a demo for users, what do you think?

morfeusys: I don't think so

Denire (approved):

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
```diff
@@ -18,11 +19,34 @@ class TelegramReactions(
 ) : Reactions() {
 
     val chatId = request.chatId
+    private val messages = mutableListOf<Message>()
```
nikvoloshin, line 22: should we really store all previous responses instead of the last one?

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
```diff
@@ -18,11 +19,34 @@ class TelegramReactions(
...
+        pair.first?.body()?.result?.let { messages.add(it) }
+    }
 
     override fun say(text: String): SayReaction {
         return sendMessage(text)
     }
 
+    override fun buttons(vararg buttons: String): ButtonsReaction {
+        messages.lastOrNull()?.let { message ->
+            val keyboard = message.replyMarkup?.inlineKeyboard?.toMutableList() ?: mutableListOf()
```
nikvoloshin, line 34: As I see, `kotlin-telegram-bot` supports only inline keyboard, but it'd be better to support basic telegram `ReplyKeyboard`
(not much related to this PR, just thoughts for the future)

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
```diff
@@ -18,11 +19,34 @@ class TelegramReactions(
...
+            keyboard.addAll(buttons.map { listOf(InlineKeyboardButton(it, callbackData = it)) })
+
+            api.editMessageReplyMarkup(
+                    chatId,
+                    message.messageId,
+                    replyMarkup = InlineKeyboardMarkup(keyboard)
+            ).also {
+                messages.remove(message)
```
nikvoloshin, line 42: Maybe change it to `messages.removeIf { it.messageId == message.messageId }`?
`Message` is a huge data class with a lot fields, nested data classes, etc.
I'd not rely on Message's `equals`

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
```diff
@@ -84,7 +109,16 @@ class TelegramReactions(
         replyToMessageId: Long? = null,
         replyMarkup: ReplyMarkup? = null
     ): ImageReaction {
-        api.sendPhoto(chatId, url, caption, parseMode, disableNotification, replyToMessageId, replyMarkup)
+        api.sendPhoto(
```
nikvoloshin, line 112: Let's put all parameters on one line for consistency with others `send*` methods

nikvoloshin (changes requested):

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
morfeusys: @nikvoloshin nope, we shouldn't. But this is the way to edit any of sent message once we've started to use editors.

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
morfeusys: Telegram allows attaching only inline buttons editing the message. The user [must pass keyboard](https://core.telegram.org/bots#keyboards) with the message to display a "basic" keyboard.

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
```diff
@@ -18,11 +19,34 @@ class TelegramReactions(
...
+                    chatId,
+                    message.messageId,
+                    replyMarkup = InlineKeyboardMarkup(keyboard)
+            ).also {
+                messages.remove(message)
+                addResponse(it)
+            }
+        }
+
+        return ButtonsReaction.create(buttons.asList())
```
nikvoloshin, line 47: If there is no previous response (i.e. `messages.lastOrNull == null`) and we are unable to send buttons, we anyway will log the `ButtonReaction`. Is this the desired behavior?

nikvoloshin (changes requested):

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
nikvoloshin: Yes, you are right, thanks

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramReactions.kt
morfeusys: Like for any other channel we cannot be sure that buttons was rendered successfully. And nevertheless we register buttons response as a content that was generated by scenario.

nikvoloshin: Default implementations of `Reactions` methods don't register reactions, they just create them. So if a channel doesn't support some kind of response (and so doesn't override the method), the reaction will not be registered. So I think this is the case.

morfeusys: `Reacion` has an internal constructor and could be created with `create` factory only. My point was about other channels - we cannot guarantee that some components (like buttons) will be rendered actually but nevertheless register the output..

nikvoloshin: Ah, yes, you are right about a constructor, I missed it. 
But, anyway, when we can't guarantee that responses will not be rendered on a 3rd party side - it's okay to register them in order to tell a user that we've done all required work on the framework side.
But in the current case, the channel knows for sure that the reaction will not be even processed somehow, so in this case, the channel should not register the reaction, otherwise, it will lead to confusion for the user.
I'd suggest creating an issue for that after merge.

channels/telegram/src/main/kotlin/com/justai/jaicf/channel/telegram/TelegramBotRequest.kt
```diff
@@ -26,30 +26,33 @@ interface TelegramBotRequest: BotRequest {
...
 ): TelegramBotRequest, EventBotRequest(
-    clientId = message.chat.id.toString(),
+    clientId = message.clientId,
     input = TelegramEvent.CONTACT
-)
\ No newline at end of file
+)
+
+internal val Message.clientId
+    get() = from?.id?.toString() ?: chat.id.toString()
```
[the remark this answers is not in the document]
morfeusys: Could you please describe these cases?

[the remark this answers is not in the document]
morfeusys: Decided to rollback to chat id

nikvoloshin (approved):

PR #114, opened 2020-11-16
JAICF<>JAICP Dialer API Integration

Denire:
This pull requests add integration with Dialer API, which provides functionality for processing outbound call campaigns.

Changes:
- add dialer api full support (redial, report, setResult)
- add ping query processing
- add dialer and jaicp channels IT
- add dialer api usage examples
- minor improvements


channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -0,0 +1,143 @@
...
+            error("startDateTime must be less than finishDateTime")
+        }
+    }
+}
+
+
+private fun checkLocalTime(data: JaicpDialerAPI.RedialData) {
+    val formatter = DateTimeFormatter.ofPattern("HH:mm")
+    val start = data.localTimeFrom?.let {
+        formatter.parse(it)
```
nikvoloshin, line 118: Why not `LocalTime.parse(str, formatter)`?

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -0,0 +1,143 @@
...
+import com.justai.jaicf.channel.jaicp.JSON
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonElement
+import java.time.DayOfWeek
+import java.time.Instant
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
+
+@Serializable
+class JaicpDialerAPI {
```
nikvoloshin, line 12: All of the fields and methods of this class are internal or private, why the class is not internal then?
Or maybe vice-versa, why all of its methods are internal? I think, there is nothing to hide

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/reactions/JaicpReactions.kt
```diff
@@ -51,8 +52,12 @@ open class JaicpReactions : Reactions() {
                 ""
             }
         }
+        val isTelephonyBot = this is TelephonyReactions
```
nikvoloshin, line 55: I think, we can inline this variable. It doesn't add much sense

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/reactions/TelephonyReactions.kt
```diff
@@ -1,25 +1,105 @@
...
 import com.justai.jaicf.reactions.Reactions
+import java.time.DayOfWeek
+import java.time.Instant
 
 val Reactions.telephony
     get() = this as? TelephonyReactions
 
 class TelephonyReactions : JaicpReactions() {
+    /**
+     * Appends audio to the response.
```
nikvoloshin, line 18: I don't think we really need this javadoc here as it's the same as for base `Reactions` and doesn't describe some new telephony-related logic

nikvoloshin (changes requested):

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
Denire: It should be internal, i think.
There's nothing to hide, but we provide full class usage from TelephonyReactions. I assume that the less public classes a module provides, the better it is for user.

nikvoloshin: I don't think we should look at the framework core only from the DSL usage side.
A user might want to operate with `RedialData` or `JaicpRedialAPI` programmatically (i.e. not just from DSL).

Denire: I'm not sure about a use case for it. All constructor fields from RedialData are exposed by `reactions.telephony?.redial`.

nikvoloshin: Yeah, but you are not able to construct `RedialData` and operate with it, do some datetime calculations, store it somewhere, etc. You are not able to operate with CallReportData, serialize it in some other format, print it, etc.
You are only able to do stuff that is explicitly exposed by Reactions, it seems like an unnecessary restriction for the user.

DSL (and any other facade) should **extend** model with some helpers, but it definitely shouldn't **hide** the model behind it.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -0,0 +1,113 @@
...
+    THU("thu"),
+    FRI("fri"),
+    SAT("sat"),
+    SUN("sun")
+}
+
+private fun checkStartFinishTime(data: JaicpDialerAPI.RedialData) {
+    val st = data.startDateTime
+    val fin = data.finishDateTime
+    if (st != null && fin != null) {
```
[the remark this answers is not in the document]
Denire: I don't think we should check for this kind of errors. We should check only most primitive and obvious.

Generally, If `fin < System.currentTimeMillis() (Plus some time for network latency and dialer schedule)`, redial won't be scheduled because redial finish time had passed. And I don't see a proper solution for this.

For example, your fin value can be correct by framework (current time + 1 min), but dialer has internal call pool and there were no place in pool for minute, therefore scheduled call won't redial.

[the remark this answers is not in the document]
nikvoloshin: That's true about really complex checks, but we can at least check that `st` and `fin` are greater than 0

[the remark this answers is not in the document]
Denire: It makes sense, but throwing this error may mislead user about logic inside dialer.

`st` and `fin` being positive numbers is not a requirement of dialerAPI. The requirement is that st/fin must be a **reasonable** timestamp, and we cannot check it, and I guess we should not throw misleading errors.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
Denire: Well, If you need it, you can ctrl-C and ctrl-V this class from source code and write one extension to TelephonyReactions :) 

I found two arguments against exposing these DTOs:
1. RedialData uses Long for timestamp and this is error prone. Reactions API uses Instant, that's better. Putting Instant to DTO requires a custom serialization logic, and I don't like it.
2. I'd like to keep reactions api as minimalistic as possible. For any extra cases there are extension functions and source code. I need a reason (a request from user, for example) to include it into api methods, and i'm not sure that this request will ever come.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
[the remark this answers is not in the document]
nikvoloshin: Well, we can say the same about `st <= fin` check, can't we?
I don't really figure out, why we put some little checks on our own and discard others?

Maybe I misunderstand you, what do you mean by saying that:
> st and fin being positive numbers is not a requirement of dialerAPI

Is it possible to pass a negative timestamp to the dialerAPI and don't get an error?
If not, then it is a requirement of dialerAPI

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
nikvoloshin: >  If you need it, you can ctrl-C and ctrl-V this class

That's definitely an unnecessary limitation as I said before.

> RedialData uses Long for timestamp and this is error prone.

We definitely can make `RedialData` public and write as many different constructors, factories and helpers as we want (and as a user wants) and not be limited by one method in Reactions class.

> I'd like to keep reactions api as minimalistic as possible.

The minimalistic view on this shouldn't limit the user. 

> For any extra cases there are extension functions and source code

What an extension I'm supposed to write if I want to operate on `RedialData` **without** actually triggering redial? What an extension I'm supposed to write if I want to access the report? 

I do not really see any motivation to prevent a user from doing that.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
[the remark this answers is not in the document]
Denire: > Well, we can say the same about st <= fin check, can't we?

Well, yes. But in this case we can provide accurate error message which will tell where user is wrong exaclty.

The only requirement of dialerAPI is that start/fin should make sense. **We only check things which are possible to check on our own**, for example, HHmm formatting, maxAttempts or retryInterval.

**What do we want to check**: that with given start/finish time we will redial in provided interval.
**How we're gonna check this:**  no idea, this particular check is impossible for us. 

So we decide to check some other thing, which does not solve our problem and does mislead user about internal logic. Given an error like `start and fin should be greater then zero` user will still have no idea why his params `start=10, fin=20` do not work.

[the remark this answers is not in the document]
nikvoloshin: Well, again, there is no difference with `st <= fin`.

> Given an error like "start and fin should be greater then zero" user will still have no idea why his params start=10, fin=20 do not work.

It will say to the user, that params `start = -5` is not correct.

The same for `st <= fin`: user will still have no idea why params `start = 10, fin = 20` do not work, but he definitely will know that `start = 20, fin = 10` do not work.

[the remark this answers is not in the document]
Denire: But difference is still present: error message for `st <= fin` is accurate and explains why user is wrong in this particular piece of code.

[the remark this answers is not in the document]
nikvoloshin: What is the problem with `Time cannot be negative`?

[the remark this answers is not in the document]
Denire: It sounds like "redial me please yesterday". It's an idiot-check, but not a check which should be in source code.

You can just mess up parameters order when invoking a reaction, and you have st < fin check for it. But do we need to check  user code like `Instant.ofEpochMillis(-5)`?

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
Denire: I forgot about internal constructors, you're right with `RedialData`. It's a complex object and there may be reasons to build it in DSL.

But im still not sure with removing `internal` modified on JaicpDialerAPI class. Just cannot find a reason someone would need it.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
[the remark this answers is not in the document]
nikvoloshin: What about `ArrayIndexOutOfBoundsException`? I'd not say that it is an idiot check.

>  But do we need to check user code like Instant.ofEpochMillis(-5)?

Yes we do, as we don't check Instant, we check Long, and it's pretty easy to forget to add such a check if we change the constructor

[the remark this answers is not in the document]
Denire: `ArrayIndexOutOfBoundsException` is quite different thing. Your array is dynamic in some way, but not time in this case.

I refuse to add this check because it conflicts with my understanding of programming and writing of a framework. If we do this, we should error-check a hundred things in our framework, which just MAY occur if user don't understand what he/she's doing. 

You can add this commit if you want, I just don't want to be an author of this check.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
nikvoloshin: `JaicpReadialApi` is actually just a response builder. So, the same as for `ReadialData`, why a user is not allowed to build a response programmatically?

Denire: Because all build functionality is exposed via reactions. Same reason why `JaicpReplies` were internal and sealed.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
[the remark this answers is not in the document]
nikvoloshin: >  Your array is dynamic in some way

In any way `array[-1]` is as much "idiotic" as `startTime = -1`

>  If we do this, we should error-check a hundred things in our framework, which just MAY occur if user don't understand what he/she's doing.

And you are absolutely right. That's how it should be done.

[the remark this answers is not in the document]
Denire: If you write in Kotlin and not Python, then yes, array[-1] is idiotic. 
If you want to schedule a redial (an operation in future) use negative time (past) is idiotic.

 > And you are absolutely right. That's how it should be done.

Do we assume that our users are cats who have no conception of `redial`?

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
nikvoloshin: > Same reason why JaicpReplies were internal and sealed.

Yeah, and now we have #113.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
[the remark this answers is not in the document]
nikvoloshin: Do we assume, that our users live in another universe, where start time could be bigger than finish time?

Don't we assume, that our users can use computable values in parameters, that can become negative because of overflow, for example?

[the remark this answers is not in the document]
Denire: What crazy mess should a man write before redial to suffer from long overflow? 
If a man writes that, no checks will help him.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -0,0 +1,163 @@
...
+        DayOfWeek.SATURDAY -> "sat"
+        DayOfWeek.SUNDAY -> "sun"
+    }
+}
+
+private fun checkStartFinishTime(data: JaicpDialerAPI.RedialData) {
+    val st = data.startDateTime
+    val fin = data.finishDateTime
+    if (st != null && fin != null) {
+        require(st >= fin) {
```
nikvoloshin, line 131: Invert condition

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -0,0 +1,163 @@
...
+            "The redial start time (startDateTime) must be less than redial finish time (finishDateTime)"
+        }
+    }
+}
+
+private fun checkLocalTime(data: JaicpDialerAPI.RedialData) {
+    val formatter = DateTimeFormatter.ofPattern("HH:mm")
+    val start = data.localTimeFrom?.let {
```
nikvoloshin, line 139: Let's write it in one line

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -0,0 +1,163 @@
...
+    if (start != null && end != null) {
+        require(start.isBefore(end)) {
+            "localTimeFrom cannot be higher then localTimeTo"
+        }
+    }
+}
+
+private fun checkRetryAndInterval(data: JaicpDialerAPI.RedialData) {
+    data.retryIntervalInMinutes?.let {
+        require(it < 1) {
```
nikvoloshin, line 154: Invert the condition

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
Denire: How exactly?

nikvoloshin: Just like that

```
val start = data.localTimeFrom?.let { LocalTime.parse(it, formatter) }
val end = data.localTimeTo?.let { LocalTime.parse(it, formatter) }
```

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -151,7 +146,7 @@ private fun checkLocalTime(data: JaicpDialerAPI.RedialData) {
 
 private fun checkRetryAndInterval(data: JaicpDialerAPI.RedialData) {
     data.retryIntervalInMinutes?.let {
-        require(it < 1) {
+        require(it > 1) {
```
nikvoloshin, line 149: Should be `it >= 1` I guess

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -128,20 +127,16 @@ private fun checkStartFinishTime(data: JaicpDialerAPI.RedialData) {
     val st = data.startDateTime
     val fin = data.finishDateTime
     if (st != null && fin != null) {
-        require(st >= fin) {
+        require(st <= fin) {
```
nikvoloshin, line 130: Should be `st < fin` I guess

nikvoloshin (approved):

morfeusys (approved):

Issue #43, opened 2020-09-08
DialogContext#transitions is always empty.


nikvoloshin:
DialogContext#transitions, that is used for strict (button?) transitions is never populated with new transitions and always remains empty.

Denire:
resolved with https://github.com/just-ai/jaicf-kotlin/pull/116

PR #115, opened 2020-11-21
#43 add strict transitions from buttons

Denire:
This pull request makes use of strict transitions with buttons. 
After it's merged, it will be possible to write:
```
reactions.buttons("Mew" to "/mew", "Wake up" to "/wakeup")
```

And when `Mew` or `Wake up` is clicked, corresponding state `/mew` or `/wakeup` will be activated.

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
@@ -2,6 +2,8 @@ package com.justai.jaicf.context
 
 import com.justai.jaicf.api.BotRequest
 import com.justai.jaicf.helpers.action.smartRandom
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
```diff
@@ -73,4 +75,11 @@ open class ActionContext(
      * @param texts a list of String to select random from
      */
     fun Reactions.sayRandom(texts: List<String>) = say(random(texts))
+
+    infix fun ButtonsReaction?.toStates(states: List<String>) {
```
Denire, line 79: Not sure about this extension, but we need some transition management for channel-specific buttons.

Denire:
@nikvoloshin suggested better solution with activators. This PR will be declined.

PR #116, opened 2020-11-22
#43 strict transitions impl with buttons

Denire:
This pull request makes use of strict transitions with buttons.
After it's merged, it will be possible to write:
```
reactions.buttons("Mew" to "/mew", "Wake up" to "/wakeup")
```
And when Mew or Wake up is clicked, corresponding state /mew or /wakeup will be activated.

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
@@ -7,6 +7,7 @@ import com.justai.jaicf.activator.selection.ActivationSelector
 import com.justai.jaicf.activator.catchall.CatchAllActivator
 import com.justai.jaicf.activator.event.BaseEventActivator
 import com.justai.jaicf.activator.intent.BaseIntentActivator
+import com.justai.jaicf.activator.strict.ButtonActivator
 import com.justai.jaicf.api.BotApi
 import com.justai.jaicf.api.BotRequest
 import com.justai.jaicf.api.hasQuery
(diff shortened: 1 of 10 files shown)


core/src/main/kotlin/com/justai/jaicf/activator/strict/ButtonActivator.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.justai.jaicf.model.scenario.ScenarioModel
+
+class ButtonActivator : Activator {
+
+    override val name = "buttonActivator"
+
+    override fun canHandle(request: BotRequest): Boolean = request.hasQuery()
+
+    override fun activate(botContext: BotContext, request: BotRequest, selector: ActivationSelector): Activation? {
+        val req = request.input.toLowerCase()
```
nikvoloshin, line 20: Why `toLowerCase`? I think that button press preserves the text case and case mismatching shows that the text was typed manually

core/src/main/kotlin/com/justai/jaicf/helpers/kotlin/Extensions.kt
```diff
@@ -0,0 +1,3 @@
+package com.justai.jaicf.helpers.kotlin
+
+internal fun <T> Boolean.runIfTrueElseNull(function: () -> T?): T? = if (this) function.invoke() else null
```
nikvoloshin, line 3: Let's think about a shorter name as there are many cases on using it, but the name is neraly as much long as function's body. Moreover this function is internal.
What about just `ifTrue`?

core/src/main/kotlin/com/justai/jaicf/BotEngine.kt
```diff
@@ -69,9 +70,11 @@ class BotEngine(
         fun MutableList<Activator>.removeIfPresent(a: Activator) = removeIf { it.name == a.name }
         fun MutableList<Activator>.pushToTheEnd(a: Activator) = find { it.name == a.name } ?: add(a)
 
-        val builtinActivators = listOf(BaseEventActivator, BaseIntentActivator, CatchAllActivator).map { it.create(model) }
+        val builtinActivators =
+            listOf(BaseEventActivator, BaseIntentActivator, CatchAllActivator).map { it.create(model) }
+        val strictActivators = mutableListOf(ButtonActivator).map { it.create(model) }
 
-        return toMutableList().apply {
+        return strictActivators + toMutableList().apply {
```
nikvoloshin, line 77: Let's better think about priority of strictActivators

nikvoloshin (changes requested):

core/src/main/kotlin/com/justai/jaicf/activator/strict/ButtonActivator.kt
Denire: As I remember, some channels may convert button input to upper or lower case

nikvoloshin (approved):

morfeusys (approved):

PR #117, opened 2020-11-24
Add multiple-intent activator

vintagentleman:
Hi!

This PR adds a public `intents` activator, which activates a state on any intent having the same prefix as the one provided.

The motivation behind this is that my classifier has a number of intents (FAQ intents in my case) that all need to be processed in the same way, e.g.
```kotlin
activators {
    intent("FAQ/Address")
    intent("FAQ/BusinessHours")
    intent("FAQ/WorkEmail")
    intent("FAQ/WorkPhone")
    // ...
}

action {
    activator.caila?.topIntent?.answer?.let {
        reactions.say(it)
    }
}
```

Adding all intents like this would be tedious and non-scalable, and I can’t use `anyIntent()` since there are other intents in my classifier that have nothing to do with FAQ.

This feature solves the problem and allows rewriting the example above simply as
```kotlin
activators {
    intents("FAQ")
}
```


vintagentleman:
Closing this as per [this discussion](https://just-ai.slack.com/archives/CTCAG8ZM2/p1606208261080700).

PR #124, opened 2020-12-02
Generate intents list in CailaIntentActivator

morfeusys:


Denire:
This PR is closed after a discussion.

The general issue (to activate by intent path, for example) will be resolved with new scenario DSL + https://github.com/just-ai/jaicf-kotlin/pull/145

This will enable writing custom activation rules in scenario with not-so-hard-to-write extension functions.

PR #128, opened 2020-12-18
Implement generic actions

nikvoloshin:
* Every activator, channel, whatever, could define its own type token that holds information about certain types of context variables. This token can be used in different contexts in order to provide type-specific behavior.
* Type tokens can be composed with each other (e.g. `telegram and caila`)
* Type tokens can be invoked with lambdas inside the action block
* Type token can be passed to the `action` function as a parameter
* ActionContext became generic

### DIFF
--- a/channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsTypeToken.kt
+++ b/channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsTypeToken.kt
@@ -6,7 +6,7 @@ import com.justai.jaicf.generic.ContextTypeToken
 
 typealias ActionsTypeToken = ChannelTypeToken<ActionsBotRequest, ActionsReactions>
 
-val googleActions: ActionsTypeToken = ChannelTypeToken()
+val actions: ActionsTypeToken = ChannelTypeToken()
 
 val ActionsTypeToken.intent get() = ContextTypeToken<ActionsDialogflowActivatorContext, ActionsIntentRequest, ActionsReactions>()
 val ActionsTypeToken.text get() = ChannelTypeToken<ActionsTextRequest, ActionsReactions>()
--- a/channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsTypeToken.kt
+++ b/channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsTypeToken.kt
@@ -1,0 +1,12 @@
+package com.justai.jaicf.channel.googleactions
+
+import com.justai.jaicf.channel.googleactions.dialogflow.ActionsDialogflowActivatorContext
+import com.justai.jaicf.generic.ChannelTypeToken
+import com.justai.jaicf.generic.ContextTypeToken
+
+typealias ActionsTypeToken = ChannelTypeToken<ActionsBotRequest, ActionsReactions>
+
+val googleActions: ActionsTypeToken = ChannelTypeToken()
+
+val ActionsTypeToken.intent get() = ContextTypeToken<ActionsDialogflowActivatorContext, ActionsIntentRequest, ActionsReactions>()
+val ActionsTypeToken.text get() = ChannelTypeToken<ActionsTextRequest, ActionsReactions>()
--- a/core/src/main/kotlin/com/justai/jaicf/generic/Compositions.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/generic/Compositions.kt
@@ -1,0 +1,42 @@
+package com.justai.jaicf.generic
+
+import com.justai.jaicf.api.BotRequest
+import com.justai.jaicf.context.ActivatorContext
+import com.justai.jaicf.reactions.Reactions
+
+// -- Type tokens' compositions
+
+private typealias Act<A> = ActivatorContextTypeToken<A>
+private typealias Ch<B, R> = ChannelTypeToken<B, R>
+private typealias Ctx<A, B, R> = ContextTypeToken<A, B, R>
+
+private typealias Req = BotRequest
+private typealias React = Reactions
+private typealias ActCtx = ActivatorContext
+
+// ---- ActivatorTypeToken and ActivatorTypeToken
+infix fun <A1: ActCtx, A2: A1> Act<A1>.and(other: Act<A2>): Act<A2> = other
+
+// ---- ActivatorTypeToken and ChannelTypeToken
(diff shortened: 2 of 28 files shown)


core/src/main/kotlin/com/justai/jaicf/generic/Compositions.kt
```diff
@@ -0,0 +1,42 @@
...
+private typealias Act<A> = ActivatorTypeToken<A>
+private typealias Ch<B, R> = ChannelTypeToken<B, R>
+private typealias Ctx<A, B, R> = ContextTypeToken<A, B, R>
+
+private typealias Req = BotRequest
+private typealias React = Reactions
+private typealias ActCtx = ActivatorContext
+
+// ---- ActivatorTypeToken and ActivatorTypeToken
+infix fun <A1: ActCtx, A2: A1> Act<A1>.and(other: Act<A2>): Act<A2> = other
```
morfeusys, line 18: Why should we have such the "same type" tokens infix `and`? Maybe it could be better to signal the user that it's prohibited operator to prevent using this construction?

core/src/main/kotlin/com/justai/jaicf/generic/TypeToken.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.justai.jaicf.reactions.Reactions
+import kotlin.reflect.KClass
+
+/**
+ * Type token that holds information about certain types of [ActivatorContext], [BotRequest] and [Reactions].
+ * Can be used in some contexts in order to provide type-specific functionality.
+ * Can be composed with other type tokens, see [and] functions.
+ */
+data class ContextTypeToken<A: ActivatorContext, B: BotRequest, R: Reactions>(
+    val activatorType: KClass<A>,
```
morfeusys, line 14: Could we use `internal` modifier here to hide these params in user's scenarios?

morfeusys (approved):
Great feature! Is it possible to make separated (or the same) `ScenarioBuilder` that could be parametrised the same way as an `action` block to forward the type information through an entire scenario?

Denire (approved):

channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsTypeToken.kt
```diff
@@ -0,0 +1,12 @@
+package com.justai.jaicf.channel.googleactions
+
+import com.justai.jaicf.channel.googleactions.dialogflow.ActionsDialogflowActivatorContext
+import com.justai.jaicf.generic.ChannelTypeToken
+import com.justai.jaicf.generic.ContextTypeToken
+
+typealias ActionsTypeToken = ChannelTypeToken<ActionsBotRequest, ActionsReactions>
+
+val googleActions: ActionsTypeToken = ChannelTypeToken()
```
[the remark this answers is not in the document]
nikvoloshin: `actions` looks good with an explicit receiver, like `request.actions` or `reactions.actions`
But without the receiver it might be confusing:
```
action {
  actions {
  
  }
}
```
or
```
action(actions) {

}
```

For me, it looks confusing, but not critical, actually. 
So I can rename this if you strongly suggest.

core/src/main/kotlin/com/justai/jaicf/generic/Compositions.kt
nikvoloshin: No problems arise by using such overloads, and I don't think it must be prohibited, it looks like a limitation without an actual reason. They just exist for consistency and do nothing.

core/src/main/kotlin/com/justai/jaicf/generic/TypeToken.kt
nikvoloshin: It might be useful to hide these parameters from scenarios, but I definitely don't want to make it internal, as it will limit us and a user in developing some new features based on these tokens.

nikvoloshin:
> Is it possible to make separated (or the same) ScenarioBuilder that could be parametrised the same way as an action block to forward the type information through an entire scenario?

Technically it is possible, but I don't think we should do that now.
For now, such a behavior can be simulated by defining a simple single-line extension function that will replace `action {}`, and I don't want to introduce a new semantic, that just replaces a single function.

But we can think better on some additional functionality we can provide with such a type-specific scenario, and if we come up with something, then we can provide such a semantic.
For example, we can greatly simplify the `activators` block, if we know the type of activator.

So I'd rather leave such type-specific scenarios semantics for future ideas.

channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsTypeToken.kt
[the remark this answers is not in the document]
morfeusys: I suggest to pick the same name for both in this case. `actions` or `googleActions`

core/src/main/kotlin/com/justai/jaicf/generic/Compositions.kt
morfeusys: Well, it looks confusing for me because it's not clear to resolve which one of two tokens will be used.

core/src/main/kotlin/com/justai/jaicf/generic/TypeToken.kt
morfeusys: What kind of features do you mean?

morfeusys:
> > Is it possible to make separated (or the same) ScenarioBuilder that could be parametrised the same way as an action block to forward the type information through an entire scenario?
> 
> Technically it is possible, but I don't think we should do that now.
> For now, such a behavior can be simulated by defining a simple single-line extension function that will replace `action {}`, and I don't want to introduce a new semantic, that just replaces a single function.
> 
> But we can think better on some additional functionality we can provide with such a type-specific scenario, and if we come up with something, then we can provide such a semantic.
> For example, we can greatly simplify the `activators` block, if we know the type of activator.
> 
> So I'd rather leave such type-specific scenarios semantics for future ideas.

Well, if there is some complex scenario, someone could have a lot of `action` blocks and such a feature will replace more than a single function.

core/src/main/kotlin/com/justai/jaicf/generic/Compositions.kt
nikvoloshin: The method's signature says that the most specific type will be used.
I can describe it in a comment before all compositions, as there's common logic for all methods: 
* the second argument always must be more (or no less) specific by all arguments than the first one
* the result is always a type-token of the most specific types

morfeusys (approved):

PR #139, opened 2021-02-02
Rewrite Scenario DSL

### LINKED ISSUE #42: Fix DSL methods nesting in Scenario DSL
Now it's possible to write Scenario like this:
```
state("") {
    activators { 
        action { 
            globalActivators { 
                
            }
        }
    }
}
```

Possible solution: use `@DslMarker` ([doc](https://kotlinlang.org/docs/reference/type-safe-builders.html)).
But it requires to rework ScenarioBuilder architecture.

### LINKED ISSUE #84: Channel-specific scenarios
Looks like it's common use-case when the user have to create dialogue scenario for only a single channel. In this case accessing a channel-specific interfaces like `reactions.alexa?` or  `request.alexa?` could be annoying.
As I think we could provide some channel-specific builders to make it possible to automatically cast `BotContext`'s interfaces to channel-specific and activator-specific implementations.

nikvoloshin:
This PR replaces the old Scenario DSL with the new one in order to support `@DslMarker`

There are no major changes in DSL functions, they are the same and work like before.

The changes worth mentioning:
- Scenarios no longer can extend `ScenarioBuilder` (or `Scenario` typealias) because of limitations of `DslMarker`
- `BotEngine` no more requires `ScenarioModel`, it requires the object of new `Scenario`  type
- `Scenario` is simply a holder of `ScenarioModel`, so users no longer need to do `MyScenario.model`
- There are multiple functions at **com.justai.jaicf.builder.Scenario.kt** that builds a scenario and returns `Scenario` object. So it's up to the user where to store `Scenario` object in order to pass it to `BotEngine`.
- `ScenarioBuilder` no longer has `dependencies` parameters, `ScenarioModel` no longer has `plus` function. Instead more explicit `append` function were added. `ScenarioBuilder#append` is used for append another scenario to the end of this one. `Scenario.append` is used to append one scenario to the end of another.
- Users are no longer allowed to define states directly on the top level of scenario building next to `append` or `handle` functions. Instead, users have to start scenario building by calling `ScenarioBuilder#start` function. However, there is **com.justai.jaicf.builder.Scenario.kt#startScenario** that allows skipping `start` function call, if a user doesn't need to add handlers or append other scenarios.
- **Channel-specific Scenarios!** (see. telephony-example)
- MANY WAYS TO MERGE SCENARIOS

More on creating scenarios:
To create a `Scenario` object a user should simply call `Scenario` or `startScenario` function.

Option 1:
```kotlin
val MyScenario = Scenario {
  // your scenario
}
BotEngine(scenario = MyScenario, ...)
```

Option 2:
```kotlin
object MyScenario {
  val scenario = Scenario {
    // your scenario
  }
}
BotEngine(scenario = MyScenario.scenario)
```

Option 3:
```kotlin
object MyScenario : Scenario {
  override val scenario by Scenario {
    // your scenario
  }
}
BotEngine(scenario = MyScenario)
```

Closes #84
Closes #42


Denire:
@nikvoloshin don't forget to change READMEs and some javadocs which still use old way of creating scenario via
```
object MainScenario: Scenario() { ... }
```

Denire:
@nikvoloshin Also, I remember you told me that we can get rid of `start` method inside `Scenario` and remove extra indentation level. How we reach first `action` block with level-5 indentation, which looks like callback-hell. 
```
object GameSetupScenario: Scenario {
    const val state = "/setup"
    override val model by Scenario {
        append(GamersCountScenario(2, supportedColors.size), GamersColorsScenario)
        start {
            state(state) {
                action {
                    val game = GameController(context)
                    if (game.gamers == null) {
                        reactions.say("Okay! Let's start a new game!")
                        reactions.go(GamersCountScenario.state, "next")
                    } else {
                        reactions.go("next")
                    }
                } // indentation hell begins
            }
        }
    }
}
...
```

Do you think it's essential to keep `start` method?

core/src/main/kotlin/com/justai/jaicf/builder/Scenario.kt
```diff
@@ -0,0 +1,36 @@
...
+    body: ScenarioBuilder<B, R>.() -> Unit
+): Scenario = createScenario { ScenarioBuilder(channelToken).apply(body).build() }
+
+@ScenarioDsl
+fun Scenario(
+    body: ScenarioBuilder<BotRequest, Reactions>.() -> Unit
+): Scenario = Scenario(ChannelTypeToken.Default, body)
+
+@ScenarioDsl
+fun <B : BotRequest, R : Reactions> startScenario(
```
Denire, line 22: Why not `createScenario` or `buildScenario`?

core/src/main/kotlin/com/justai/jaicf/model/scenario/Scenario.kt
```diff
@@ -1,9 +1,14 @@
...
- * @see ScenarioBuilder
- */
-typealias Scenario = ScenarioBuilder
\ No newline at end of file
+interface Scenario : ReadOnlyProperty<Any, ScenarioModel> {
+    val model: ScenarioModel
+    override fun getValue(thisRef: Any, property: KProperty<*>) = model
+}
+
+fun createScenario(model: () -> ScenarioModel): Scenario = object : Scenario {
```
Denire, line 12: I think this method can be internal

Denire (changes requested):

core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
```diff
@@ -86,30 +87,13 @@ open class ActionContext<A: ActivatorContext, B: BotRequest, R: Reactions>(
...
-    operator fun <A1: A, B1: B, R1: R, T> ContextTypeToken<A1, B1, R1>.invoke(action: ActionContext<A1, B1, R1>.() -> T): T? {
-        return if (isInstance(activator)) {
-            @Suppress("UNCHECKED_CAST")
-            (this@ActionContext as ActionContext<A1, B1, R1>).action()
-        } else {
-            null
-        }
-    }
+    operator fun <B1: B, R1: R, T> ChannelTypeToken<B1, R1>.invoke(action: ActionContext<A, B1, R1>.() -> T): T? =
+        safeCast(this)?.run(action)
```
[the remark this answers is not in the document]
nikvoloshin: It's reasonable, but I think we need logging only in `action(token)` and not in `invoke`, as it's intended behavior for `invoke` to skip block if a token doesn't match while in `action` block it really can indicate mistakes.

core/src/main/kotlin/com/justai/jaicf/builder/Scenario.kt
nikvoloshin: To indicate that it's related to `start` function

core/src/main/kotlin/com/justai/jaicf/context/ActionContext.kt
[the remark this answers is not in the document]
Denire: Trace-level logs then maybe? 

The reason why i'd like some logs to be in application is that it's really easy for new user to forget or even don't know about casting behaviour. So client action block won't be invoked and there will be no way to understand why, except debugging or reading wiki (small **NOTE:** block in the footer of the action wiki page).

I think framework should tell about request execution in logs, but not hide it behind wiki pages.

core/src/main/kotlin/com/justai/jaicf/builder/Scenario.kt
Denire: Since method `createScenario` became internal, I think we can reconsider `startScenario` naming, and `start` method itself.
Even if we still keep `start` method, I think naming is incorrect. We don't start anything, we don't launch any processing, we're `creating` or `building` a scenario.

morfeusys (approved):
The only change I'd like to see is simplified scenarios in hello-world example that doesn't use `Scenario` delegation and also uses activators' type tokens inside actions.

Denire (approved):

PR #144, opened 2021-02-27
feature/before-activation-hook

Denire:
implement before-activation-hook

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/api/BotRequest.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/api/BotRequest.kt
@@ -28,9 +28,26 @@ enum class BotRequestType {
  */
 open class QueryBotRequest(
     override val clientId: String,
+    input: String
+) : BotRequest {
+
+    private var inputValue = input
+
     override val input: String
-): BotRequest {
-    override val type = BotRequestType.QUERY
+        get() = inputValue
+
+    internal fun setInput(input: String) {
+        inputValue = input
+    }
+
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/com/justai/jaicf/api/BotRequest.kt
```diff
@@ -28,9 +28,26 @@ enum class BotRequestType {
  */
 open class QueryBotRequest(
     override val clientId: String,
+    input: String
+) : BotRequest {
+
+    private var inputValue = input
```
nikvoloshin, line 34: Why not var with private set?

core/src/main/kotlin/com/justai/jaicf/api/BotRequest.kt
```diff
@@ -28,9 +28,26 @@ enum class BotRequestType {
...
     override val input: String
-): BotRequest {
-    override val type = BotRequestType.QUERY
+        get() = inputValue
+
+    internal fun setInput(input: String) {
+        inputValue = input
+    }
+
+    private var typeValue = BotRequestType.QUERY
```
nikvoloshin, line 43: Why not var with private set?

nikvoloshin (approved):

core/src/main/kotlin/com/justai/jaicf/api/BotRequest.kt
Denire: private setters are not allowed for open properties

morfeusys (approved):

PR #166, opened 2021-04-12
move EndSessionReaction to core

veptechno:
Many channels can end session. But end session reaction is locked inside JAICP module. This moves EndSessionReaction to core and unifies the reaction to all channels.


channels/alexa/src/main/kotlin/com/justai/jaicf/channel/alexa/AlexaReactions.kt
```diff
@@ -97,11 +98,14 @@ class AlexaReactions(
         response.builder.addAudioPlayerStopDirective()
     }
 
-    fun endSession(text: String? = null) {
-        text?.let { say(it) }
+    fun endSession(text: String): EndSessionReaction {
```
morfeusys, line 101: Why did you change the method's signature? `text` arg should be nullable for multi-channel scenarios where some content could be generated for many channels before the session ending.

morfeusys (changes requested):

channels/alexa/src/main/kotlin/com/justai/jaicf/channel/alexa/AlexaReactions.kt
veptechno: Ok

Denire:
I suggest also moving `NewSessionReaction` from JAICP package somewhere close to `EndSessionReaction` as they belong close to each other.

Denire (approved):

PR #168, opened 2021-04-15
add logging in facebook messenger

veptechno:
Add logging in facebook messenger, so that framework users can understand what happens if the proxy returns an error.

### DIFF
--- a/channels/facebook/src/main/java/com/justai/jaicf/channel/facebook/messenger/Messenger.java
+++ b/channels/facebook/src/main/java/com/justai/jaicf/channel/facebook/messenger/Messenger.java
@@ -221,16 +221,21 @@ public final class Messenger {
       final Optional<String> jsonBody = payload.map(this.gson::toJson);
       final HttpResponse httpResponse =
           this.httpClient.execute(httpMethod, requestUrl, jsonBody.orElse(null));
-      final JsonObject responseJsonObject =
-          this.jsonParser.parse(httpResponse.body()).getAsJsonObject();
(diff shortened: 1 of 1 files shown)


channels/facebook/src/main/java/com/justai/jaicf/channel/facebook/messenger/Messenger.java
```diff
@@ -221,16 +221,21 @@ public SetupResponse deleteSettings(
...
+          throw new MessengerApiException(
+              "The response JSON does not contain any key/value pair", empty(), empty(), empty());
+        }
+        if (httpResponse.statusCode() >= 200 && httpResponse.statusCode() < 300) {
+          return responseTransformer.apply(responseJsonObject);
+        } else {
+          throw MessengerApiExceptionFactory.create(responseJsonObject);
+        }
+      } catch (Exception e) {
+        log.error("Http response status: {}, body: {}", httpResponse.statusCode(), httpResponse.body());
```
Denire, line 237: We'd better to indicate that an error happened in message.
I'd suggest
```
log.error("Invalid HTTP response with status: {}, and body: {}", ..)
```

Denire (changes requested):

channels/facebook/src/main/java/com/justai/jaicf/channel/facebook/messenger/Messenger.java
veptechno: Fix

Denire (approved):

PR #174, opened 2021-05-12
Implement viber sdk and channel

veptechno:
This adds a new sdk for Viber and a channel for using Viber in jaicf bots.


channels/viber/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+
+plugins {
+    `jaicf-kotlin`
+    `jaicf-publish`
+}
+
+dependencies {
+    core()
+    api(`coroutines-core`())
+    implementation(ktor("ktor-client-cio"))
```
Denire, line 15: I think these dependencies should be API level

channels/viber/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+    implementation(ktor("ktor-client-logging-jvm"))
+    implementation(ktor("ktor-server-netty"))
```
Denire, line 17: why do we need netty?

examples/hello-world/src/main/kotlin/com/justai/jaicf/examples/helloworld/channel/viber/scenario/ViberTestScenario.kt
```diff
@@ -0,0 +1,277 @@
...
+import com.justai.jaicf.channel.viber.sdk.message.PictureMessage
+import com.justai.jaicf.channel.viber.seen
+import com.justai.jaicf.channel.viber.sticker
+import com.justai.jaicf.channel.viber.subscribed
+import com.justai.jaicf.channel.viber.text
+import com.justai.jaicf.channel.viber.viber
+import com.justai.jaicf.channel.viber.video
+import java.io.File
+
+val ViberTestScenario = Scenario {
```
Denire, line 30: I don't think this scenario should exist in Hello-world example. But the code is useful, so I'd suggest moving it to separate module, like `viber-sdk-example`.

Denire (changes requested):

channels/viber/build.gradle.kts
veptechno: Ok

channels/viber/build.gradle.kts
veptechno: Forgot to remove

examples/hello-world/src/main/kotlin/com/justai/jaicf/examples/helloworld/channel/viber/scenario/ViberTestScenario.kt
veptechno: You're right, I created an additional package for this. I also added a config class that can be used to specify to ignore seen and delivered events, so the viber channel can now be used in the default HelloWorldScenario.

nikvoloshin:
@veptechno it'd be nice if you refactor code converted from Java and make it look more Kotlin-like where possible. For instance, get rid of streams and other Java-specific stuff.

Denire (approved):

channels/viber/src/main/kotlin/com/justai/jaicf/channel/viber/sdk/Jackson.kt
```diff
@@ -0,0 +1,18 @@
+package com.justai.jaicf.channel.viber.sdk
```
Denire, line 1: This file should have proper JvmName for java clients. Otherwise generated name will be JacksonKt, which is a bad name for public class

PR #178, opened 2021-05-26
Fix google-actions reactions

veptechno:
The reactions of the google-actions channel worked with unexpected behavior. Implemented more fail-safe behavior


channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsReactions.kt
```diff
@@ -34,7 +43,12 @@ class ActionsReactions(
     private fun clean(text: String) = text.replace(ssmlRegex, " ")
 
     override fun say(text: String): SayReaction {
-        addSimpleResponse(clean(text), text)
+        val simpleResponse = SimpleResponse().also {
```
Denire, line 46: SimpleResponse has setters, we don't have to use also here

channels/google-actions/src/main/kotlin/com/justai/jaicf/channel/googleactions/ActionsReactions.kt
```diff
@@ -47,29 +61,19 @@ class ActionsReactions(
...
+        val image = Image().setUrl(url)
+        accessibilityText?.let(image::setAccessibilityText)
+
+        response.builder.add(image)
         return ImageReaction.create(url)
     }
 
     fun endConversation() = response.builder.endConversation()
 
-    fun addSimpleResponse(displayText: String, ssml: String = displayText) {
```
Denire, line 58: Deleting this public function breaks compatibility with previous versions. We should remain this signature.

Denire (changes requested):

Denire (approved):

PR #180, opened 2021-06-02
Update CailaInferenceResultData.kt

discodasha:
update CAILA format for 1.10.11 release


Denire:
I don't think we should add priority to our data model. This is internal fields for CAILA and there is no use in adding them.

PR #184, opened 2021-07-15
Fix state path parsing and state name verification

veptechno:
This PR fixes parsing of multiple slashes in go, changeState and other reactions. 
It also prohibits slashes in state names (except for one leading slash in top-level states).

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
@@ -305,7 +305,7 @@ class StateBuilder<B : BotRequest, R : Reactions> internal constructor(
     internal override fun build(): State = verify().run { State(path, noContext, modal, action?.let(::ActionAdapter)) }
 
     private fun verify(): StateBuilder<B, R> {
-        if (this.parent.isRoot && !name.matches(Regex("/?[^/]+")))
+        if (this.parent.isRoot && !name.matches(Regex("/?[^/]*")))
             logger.warn(
                 """
                     Slashes are not allowed in the name of top-level state. Your state path: "$path"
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
```diff
@@ -291,15 +292,40 @@ class StateBuilder<B : BotRequest, R : Reactions> internal constructor(
...
         body: @ScenarioDsl ActionContext<A1, B1, R1>.() -> Unit
     ) = action { contextToken.invoke(body) }
 
-    override internal fun build(): State = State(path, noContext, modal, action?.let(::ActionAdapter))
-}
\ No newline at end of file
+    internal override fun build(): State = verify().run { State(path, noContext, modal, action?.let(::ActionAdapter)) }
+
+    private fun verify(): StateBuilder<B, R> {
+        if (this.parent.isRoot && !name.matches(Regex("/?[^/]+")))
```
nikvoloshin, line 308: This condition is true for top-level states with empty names.

core/src/main/kotlin/com/justai/jaicf/builder/ScenarioBuilder.kt
```diff
@@ -291,15 +292,40 @@ class StateBuilder<B : BotRequest, R : Reactions> internal constructor(
...
+            logger.warn(
+                """
+                    Slashes are not allowed in the name of top-level state. Your state path: "$path"
+                    Example solution: replace "state("state/child") with two separate states.
+                    This may cause incorrect JAICF Intellij IDEA Plugin behaviour.
+                    """.trimIndent()
+            )
+
+        if (!this.parent.isRoot && !name.matches(Regex("[^/]+")))
```
nikvoloshin, line 317: This condition is true for inner states with empty names.

nikvoloshin (approved):

Denire (approved):

PR #189, opened 2021-07-27
Bot Routing Feature

Denire:
This PR brings Bot Routing feature to switch executable `BotEngines` in runtime. 

ChangeSet:
1. New `BotEngine` implementation - `BotRoutingEngine`. This entity works with multiple BotEngines, preserves dialogContext for each one and changes current `BotEngine` for client.
2. `BotRoutingApi` - an API with routing methods (route, routeBack, changeEngine, changeEngineBack) available from `action` context via `routing` extension, e.g.:
```kotlin
action {
  routing.route("another-engine")
}
```

Usage example can be found in [multilingual bot example](https://github.com/just-ai/jaicf-kotlin/tree/bot-routing-api/examples/multilingual-bot/src/main/kotlin/com/justai/jaicf/examples/multilingual).


veptechno (approved):

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @param main main engine to process requests.
+ * @param routables a map of engines with names to perform route from and to.
+ *
+ * @see BotRoutingApi scenario runtime api for routing.
+ * */
+class BotRoutingEngine(
+    private val main: BotEngine,
+    routables: Map<String, BotEngine>,
+) : BotEngine(MOCK_SCENARIO), WithLogger {
```
morfeusys, line 57: Looks like `BotRoutingEngine` is only a decorator for multiple BotEngines.
It's not clear for me why we need `BotEngine` implementation here, not `BotApi`?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    private fun getCurrentRouting(ctx: BotContext): String? = try {
+        ctx.routingContext.routingEngineStack.peek()
+    } catch (e: EmptyStackException) {
+        null
+    }
+
+    companion object {
+        internal const val DEFAULT_ROUTE_NAME = "main"
```
morfeusys, line 148: Are you sure there is no way to get rid of hardcoded "main" engine name?
For example storing it with `null` key and making it possible to route to it with some explicit method like `routeToMainEngine`?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
Denire: @morfeusys 
1. BotAPI has no knowledge of contextManagers, but here we must use it.
2. BotAPI is impossible to use in BotTest (mostly because of contextManagers).

morfeusys: 1. Context manager is a part of `process()`, isn't it enough?
2. Maybe it's a time to make it possible to use BotApi in tests?

Denire: ContextManager has no part in process, only BotContext is part of process.
This can be used only if we make `BotContextManager` part of BotApi interface

morfeusys: [BotApi.process()](https://github.com/just-ai/jaicf-kotlin/blob/master/core/src/main/kotlin/com/justai/jaicf/api/BotApi.kt#L33)

Denire: @morfeusys I've edited my comment before you've seen it. But still BotContextManager is nullable, because it's channel-provided. And we cannot use it until it's nullable.

I'd suggest adding defaultContextManager to BotApi.process() if we must support BotApi.
```kotlin
interface BotApi {
    val defaultContextManager: BotContextManager
    fun process(...){}
``

Comment (morfeusys) [reply core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt:57]: Each `BotEngine` has it's own `defaultContextManager` that can be retrieved. You can add additional context manager for `BotRoutingEngine` if you need it.

File: core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt [hunk 3]
```diff
@@ -0,0 +1,164 @@
...
+ * An exception thrown to re-route current BotRequest to specified [targetEngineName]
+ *
+ * @param [targetEngineName] an engine with specified name provided in [BotRoutingEngine] routables map.
+ * */
+internal data class BotRequestRerouteException(val targetEngineName: String) : RuntimeException()
+
+/**
+ * An exception thrown when there is no previous engine to route back.
+ * */
+internal object NoRouteBackException : RuntimeException()
```
nikvoloshin, line 21: 1. Same about internal
2. Let's not make a singleton what is not a singleton in semantics

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+import com.justai.jaicf.context.DialogContext
+import com.justai.jaicf.helpers.logging.WithLogger
+import com.justai.jaicf.model.state.StatePath
+import java.util.EmptyStackException
+import java.util.Stack
+
+/**
+ * An exception thrown to re-route current BotRequest to specified [targetEngineName]
+ *
+ * @param [targetEngineName] an engine with specified name provided in [BotRoutingEngine] routables map.
```
nikvoloshin, line 14: "... provided in [BotRoutingEngine.routables] map."
Let's add a link to rouatables

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+)
+
+/**
+ * A helpful extension to get [BotRoutingContext] from BotEngine.
+ *
+ * @see BotRoutingContext
+ * @see BotRoutingApi
+ * @see BotRoutingEngine
+ * */
+internal val BotContext.routingContext: BotRoutingContext
```
nikvoloshin, line 151: Why internal?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+/**
+ * A context used in [BotRoutingEngine] and [BotRoutingApi] to store routing state.
+ *
+ * @see BotRoutingApi
+ * @see BotRoutingEngine
+ * */
+internal data class BotRoutingContext(
```
nikvoloshin, line 137: Why internal?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+     * @see BotRoutingEngine
+     * */
+    fun changeEngineBack() {
+        try {
+            val routingContext = botContext.routingContext
+            val curr = routingContext.routingEngineStack.pop()
+            val target = routingContext.routingEngineStack.peek()
+            logger.info("Changing execution back from engine: $curr to engine: $target")
+        } catch (e: EmptyStackException) {
+            logger.warn("Failed to change bot engine back as there is no engines left in stack")
```
nikvoloshin, line 126: Shouldn't we throw NoRouteBackException?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @see BotRoutingEngine
+     * */
+    fun changeEngineBack() {
+        try {
+            val routingContext = botContext.routingContext
+            val curr = routingContext.routingEngineStack.pop()
+            val target = routingContext.routingEngineStack.peek()
+            logger.info("Changing execution back from engine: $curr to engine: $target")
+        } catch (e: EmptyStackException) {
```
nikvoloshin, line 125: Shouldn't we restore context in case of an error? For example, put `curr` back to stack

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+     * @see BotRoutingEngine
+     * */
+    fun routeBack(): Nothing {
+        try {
+            val routingContext = botContext.routingContext
+            val curr = routingContext.routingEngineStack.pop()
+            val target = routingContext.routingEngineStack.pop()
+            logger.info("Routing request back from engine: $curr to engine: $target")
+            throw BotRequestRerouteException(target)
+        } catch (e: EmptyStackException) {
```
nikvoloshin, line 89: The same about restoring context here

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * A context used in [BotRoutingEngine] and [BotRoutingApi] to store routing state.
+ *
+ * @see BotRoutingApi
+ * @see BotRoutingEngine
+ * */
+internal data class BotRoutingContext(
+    val dialogContextMap: MutableMap<String, DialogContext> = mutableMapOf(),
+    val routingEngineStack: Stack<String> = Stack(),
```
nikvoloshin, line 139: Why stack? We use ArrayDeque everywhere

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,151 @@
...
+        contextManager: BotContextManager?,
+    ) {
+        val botContext = (contextManager ?: main.defaultContextManager).loadContext(request, requestContext)
+        val route = getCurrentRouting(botContext) ?: DEFAULT_ROUTE_NAME
+        processWithRouting(request, reactions, requestContext, contextManager, botContext, route)
+    }
+
+    private fun getCurrentRouting(ctx: BotContext): String? = try {
+        ctx.routingContext.routingEngineStack.peek()
+    } catch (e: EmptyStackException) {
```
nikvoloshin, line 143: Let's not use exceptions for control flow, use empty check instead

nikvoloshin (commented):
Looks good, thank you!
A single complaint: why is everything private and internal? We leave absolutely no extension points for a user. Neither BotRoutingEngine can be customized nor user can implement its own api operating on our objects.

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
Denire: General user should not see it. Anyone who wants to modify this context can copy-paste deserialization logic form sources.

nikvoloshin: A user may want to read it to build its own logic. You can make public getters and internal or private mutators

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
Denire: I don't like to store it with null key, so i've changed default key to prevent conflicts on clients using name "main" in routables.

nikvoloshin (approved):

morfeusys (approved):

Denire:
@nikvoloshin @veptechno there were changes since you last reviewed:
1. `BotRoutingEngine`s can form tree with children and parents. Therefore hierarchical routing is available.
2. Added ErrorHandler for `BotRouting`.

veptechno (approved):

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,198 @@
...
+
+    init {
+        if (routables.isEmpty()) {
+            error("Collection of routable engines must not be empty")
+        }
+        if (routables.values.distinctBy { it.defaultContextManager }.size > 1) {
+            error("Collections of routable botEngines must have single shared context manager instance")
+        }
+        if (main.second is BotRoutingEngine) {
+            error("Main bot cannot be an instance of BotRoutingEngine")
```
nikvoloshin, line 73: Why?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,198 @@
...
+        val route = getCurrentRouting(botContext) ?: RoutingRequest(routerName, routerName)
+        processWithRouting(request, reactions, requestContext, contextManager, botContext, route)
+    }
+
+    private fun getCurrentRouting(ctx: BotContext): RoutingRequest? = ctx.routingContext.routingStack.peek()
+
+    private fun getChildDescriptor(router: String): BotRoutingEngine? =
+        children.find { it.routerName == router }
+
+    private fun getParentDescriptor(): BotRoutingEngine? = parent
```
nikvoloshin, line 152: Don't think we really need this method

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,198 @@
...
+    ) {
+        val botContext =
+            (contextManager ?: routerDefaultEngine.defaultContextManager).loadContext(request, requestContext)
+        val route = getCurrentRouting(botContext) ?: RoutingRequest(routerName, routerName)
+        processWithRouting(request, reactions, requestContext, contextManager, botContext, route)
+    }
+
+    private fun getCurrentRouting(ctx: BotContext): RoutingRequest? = ctx.routingContext.routingStack.peek()
+
+    private fun getChildDescriptor(router: String): BotRoutingEngine? =
```
nikvoloshin, line 149: Don't think we really need this method

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,198 @@
...
+        executionContext0: ExecutionContext? = null,
+    ) {
+
+        val engineName = routeRequest.toEngine
+        val lastRouter = routeRequest.fromRouter ?: routerName
+        logger.info("Process route request: ${request.input} with routing for engine: $engineName and router: $lastRouter")
+
+        var router = getRouterRecursive(lastRouter) ?: this
+        if (engineName in router.childrenNames) {
+            router = getChildDescriptor(engineName) ?: router
```
nikvoloshin, line 102: if-condition implies getChildDescriptor != null, the same about else-if-condition
So we can simplify it like that, for example:
```kotlin
if (engineName in router.childrenNames) {
    router = children.first { it.routerName == engineName }
} else if (engineName == router.parent?.routerName) {
    router = router.parent!!
}
```

and get rid of get*Descriptor() functions

nikvoloshin: UPD: above is true for `router = router.getChildDescriptor(engineName)`. Looks like a mistake in the code, doesn't it?

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingApi.kt
```diff
@@ -0,0 +1,196 @@
...
+     * @see BotRoutingEngine
+     * */
+    fun routeBack(): Nothing {
+        val routingContext = botContext.routingContext
+        try {
+            val curr = routingContext.routingStack.pop()
+            val target = routingContext.routingStack.pop()
+            logger.info("Routing request back from engine: $curr to engine: $target")
+            throw BotRequestRerouteException(target)
+        } catch (e: NoSuchElementException) {
```
nikvoloshin, line 88: Let's use empty-check instead

nikvoloshin: Also, we have `hasPreviousEngineInStack` function for easy check

nikvoloshin (changes requested):

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,198 @@
...
+
+    private fun getChildDescriptor(router: String): BotRoutingEngine? =
+        children.find { it.routerName == router }
+
+    private fun getParentDescriptor(): BotRoutingEngine? = parent
+
+    private fun getRouterRecursive(router: String): BotRoutingEngine? {
+        if (router == routerName) return this
+        children.forEach { child ->
+            if (child.routerName == router) return child
```
nikvoloshin, line 157: looks like redundant check because of a first line in recursive call

core/src/main/kotlin/com/justai/jaicf/api/routing/BotRoutingEngine.kt
```diff
@@ -0,0 +1,188 @@
...
+        executionContext0: ExecutionContext? = null,
+    ) {
+
+        val engineName = routeRequest.toEngine
+        val lastRouter = routeRequest.fromRouter ?: routerName
+        logger.info("Process route request: ${request.input} with routing for engine: $engineName and router: $lastRouter")
+
+        var router = getRouterRecursive(lastRouter) ?: this
+        if (engineName in router.childrenNames) {
+            router = children.first { it.routerName == engineName }
```
nikvoloshin, line 98: We use `router.childrenNames` in if-condition,  but `this.children.first {}` inside if-block. I think one of this line should be fixed

nikvoloshin (approved):

PR #196, opened 2021-08-05
Fix modal state activation

nikvoloshin:
Fixes a bug due to which modal state itself cannot be activated from its subtree.

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/activator/BaseActivator.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/activator/BaseActivator.kt
@@ -79,28 +79,23 @@ abstract class BaseActivator(private val model: ScenarioModel) : Activator {
         }
 
     private fun generateTransitions(botContext: BotContext): List<Transition> {
-        val currentState = botContext.dialogContext.currentContext
-        val currentPath = StatePath.parse(currentState)
+        val currentPath = StatePath.parse(botContext.dialogContext.currentContext)
 
         val allStatesBack = listOf(currentPath.toString()) + currentPath.parents.reversedArray()
 
-        val availableStates = mutableListOf<String>()
-        for (state in allStatesBack) {
-            availableStates += state
+        val transitionsFrom = model.transitions.groupBy { it.fromState }
+        val availableTransitions = mutableListOf<Transition>()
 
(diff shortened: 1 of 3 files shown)


core/src/test/kotlin/com/justai/jaicf/core/test/scenario/ScdnarioModalityTest.kt
```diff
@@ -0,0 +1,68 @@
+package com.justai.jaicf.core.test.scenario
```
Denire, line 1: This file has typo error

nikvoloshin: Oops

core/src/main/kotlin/com/justai/jaicf/activator/BaseActivator.kt
```diff
@@ -94,6 +95,12 @@ abstract class BaseActivator(private val model: ScenarioModel) : Activator {
         }
 
         val transitionsMap = model.transitions.groupBy { it.fromState }
-        return availableStates.flatMap { transitionsMap[it] ?: emptyList() }
+
+        val lastState = StatePath.parse(availableStates.last())
```
Denire, line 99: Isn't it true that this lastState is always equal to currentState?

nikvoloshin: No, we process states in reverse order from current state:
Suppose currentState = /a/b/c, then the order is following: /a/b/c -> /a/b -> /a -> /
So, lastState - is either first modal ancestor, or root

Denire (approved):

core/src/main/kotlin/com/justai/jaicf/activator/BaseActivator.kt
```diff
@@ -94,6 +95,12 @@ abstract class BaseActivator(private val model: ScenarioModel) : Activator {
         }
 
         val transitionsMap = model.transitions.groupBy { it.fromState }
```
veptechno, line 97: Do you want to simplify the method?

morfeusys (approved):

PR #198, opened 2021-08-06
Add allowedTime field in RedialData

veptechno:
The localTimeFrom and localTimeTo fields in RedialData are deprecated. This PR adds a new field allowedTime and deprecates  the usages of the old format.

### DIFF
--- a/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
+++ b/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
@@ -160,6 +160,7 @@ class JaicpDialerAPI {
  */
 @Serializable
 data class AllowedTime(
+    val default: List<LocalTimeInterval>? = null,
     val mon: List<LocalTimeInterval>? = null,
     val tue: List<LocalTimeInterval>? = null,
     val wed: List<LocalTimeInterval>? = null,
@@ -167,7 +168,6 @@ data class AllowedTime(
     val fri: List<LocalTimeInterval>? = null,
     val sat: List<LocalTimeInterval>? = null,
     val sun: List<LocalTimeInterval>? = null,
-    val default: List<LocalTimeInterval>? = null
 ) {
     internal val intervals = listOf(mon, tue, wed, thu, fri, sat, sun, default)
 }
--- a/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
+++ b/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
@@ -179,10 +179,11 @@ data class AllowedTime(
     val thu: List<LocalTimeInterval>? = null,
     val fri: List<LocalTimeInterval>? = null,
     val sat: List<LocalTimeInterval>? = null,
-    val sun: List<LocalTimeInterval>? = null,
-) {
-    internal val intervals = listOf(mon, tue, wed, thu, fri, sat, sun, default)
-}
+    val sun: List<LocalTimeInterval>? = null
+)
+
+internal val AllowedTime.intervals
+    get() = listOf(default, mon, tue, wed, thu, fri, sat, sun)
 
 
 /**
@@ -225,7 +226,7 @@ private fun checkLocalTime(data: JaicpDialerAPI.RedialData) {
     val end = data.localTimeTo?.let { LocalTime.parse(it, formatter) }
     if (start != null && end != null) {
         require(start.isBefore(end)) {
-            "localTimeFrom cannot be higher then localTimeTo"
+            "localTimeFrom must be before localTimeTo"
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Denire:
@veptechno  Name of this pull request does not tell what exactly does it do. Please, rename it property for us to have a cleaner changeset and PR history

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -105,6 +146,45 @@ class JaicpDialerAPI {
...
+/**
+ * Local time interval attempting to redial
+ *
+ * @param localTimeFrom local time interval start attempting to redial. E.g. 16:20
+ * @param localTimeTo local time interval end attempting to redial. E.g. 23:59
+ * */
+@Serializable
+data class LocalTimeInterval(
+    val localTimeFrom: String,
+    val localTimeTo: String
```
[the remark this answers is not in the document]
veptechno: No, now it's impossible

Denire:
@veptechno could you please add tests in JAICP channel package for this feature?

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -105,6 +146,45 @@ class JaicpDialerAPI {
...
+    val default: List<LocalTimeInterval>? = null,
+    val mon: List<LocalTimeInterval>? = null,
+    val tue: List<LocalTimeInterval>? = null,
+    val wed: List<LocalTimeInterval>? = null,
+    val thu: List<LocalTimeInterval>? = null,
+    val fri: List<LocalTimeInterval>? = null,
+    val sat: List<LocalTimeInterval>? = null,
+    val sun: List<LocalTimeInterval>? = null,
+) {
+    internal val intervals = listOf(mon, tue, wed, thu, fri, sat, sun, default)
```
nikvoloshin, line 172: Do we really need it as a member property? Does it need to take part in a serialization process?
I think it can easily be an extension property. Furthermore, it is used in a single place, can we then completely get rid of this?

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -138,6 +218,21 @@ private fun checkLocalTime(data: JaicpDialerAPI.RedialData) {
...
+    val formatter = DateTimeFormatter.ofPattern("HH:mm")
+
+    data.allowedTime?.intervals?.filterNotNull()
+        ?.flatten()
+        ?.forEach { interval ->
+            val start = interval.localTimeFrom.let { LocalTime.parse(it, formatter) }
+            val end = interval.localTimeTo.let { LocalTime.parse(it, formatter) }
+
+            require(start.isBefore(end)) {
+                "localTimeFrom cannot be higher then localTimeTo of allowedTime"
```
nikvoloshin, line 231: `checkAllowedTime` will be in stacktrace, so I don't think we really should reference it in the error message. Also, I'd suggest changing the "lower-higher" relationship to "before-after" as we say about time.
```suggestion
                "localTimeFrom must be before localTimeTo"
```

nikvoloshin (changes requested):

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
veptechno: I added this property inside the class because if the API changes, we might forget to change this list. But it really shouldn't serialize, so it's better to put it near the class.

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
```diff
@@ -32,30 +32,35 @@ class JaicpDialerAPI {
...
         val localTimeFrom: String? = null,
+        @Deprecated("Field 'localTimeTo' are deprecated, use 'allowedTime' instead")
         val localTimeTo: String? = null,
         val maxAttempts: Int? = null,
-        val retryIntervalInMinutes: Int? = null
+        val retryIntervalInMinutes: Int? = null,
+        val allowedTime: AllowedTime? = null,
     ) {
         companion object {
+            @Deprecated("Parameters 'localTimeFrom' and 'localTimeTo' are deprecated, use 'create' method that accepts allowedTime")
```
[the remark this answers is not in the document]
veptechno: Ok

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
veptechno: It's better

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/dto/JaicpDialerAPI.kt
nikvoloshin: Yeah, I see your point, then let's keep it but as an extension function.

nikvoloshin (approved):

Denire (approved):

PR #199, opened 2021-08-09
Wrap HttpBotRequest#stream to BufferedInputStream.

### LINKED ISSUE #175: Safe wrappers for HttpBotRequest content
HttpBotRequest uses read-once streams which are not so safe for public api. An error can easily happen if someone would want to log the content before it is passed next to BotChannel.

This issue suggests wrapping HttpBotRequest streams content to read it multiple times.

nikvoloshin:
Wrap `HttpBotRequest#stream` to `BufferedInputStream` as it supports `mark` and `reset`, allowing us to read request data multiple times

Wrap `receiveText()` to `mark` + `reset` to automatically reset stream for further readings

Closes #175

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/channel/http/HttpBotRequest.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/channel/http/HttpBotRequest.kt
@@ -25,6 +25,9 @@ class HttpBotRequest(
     fun firstHeader(name: String) = headers[name]?.first()
 
     fun firstParameter(name: String) = parameters[name]?.first()
+
+    override fun toString() =
+        "HttpBotRequest(stream=$stream, headers=$headers, parameters=$parameters, requestMetadata=$requestMetadata"
 }
 
 fun <R> InputStream.runAndReset(action: InputStream.() -> R): R {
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/com/justai/jaicf/channel/http/HttpBotRequest.kt
```diff
@@ -12,20 +12,26 @@ import java.nio.charset.Charset
  * @property parameters HTTP query parameters
  * @property requestMetadata optional metadata for request processing
  */
-data class HttpBotRequest(
```
veptechno, line 15: If you remove the data modifier, then add toString and equals/hashCode methods. For example, HttpBotChannelServlet:29 uses toString.

veptechno (changes requested):

core/src/main/kotlin/com/justai/jaicf/channel/http/HttpBotRequest.kt
nikvoloshin: I added toString for logging purposes, but equals/hashCode are not really needed. It should not be a data class by its semantics.

veptechno (approved):

Denire (approved):

PR #200, opened 2021-08-11
Use GitHub actions for CI

veptechno:
Travis CI no longer works in this repository. 
This PR adds support for github actions and minor refactoring of MongoBotContextManagerTest to support an external database.


morfeusys (approved):

Denire (approved):

managers/mongo/src/test/kotlin/com/justai/jaicf/context/manager/test/MongoBotContextManagerTest.kt
```diff
@@ -75,4 +61,40 @@ class MongoBotContextManagerTest {
         assertTrue(result.result is CustomValue)
         assertTrue(result.session["value"] is CustomValue)
     }
-}
\ No newline at end of file
+}
+
+@EnabledIfEnvironmentVariable(named = "mongo", matches = "external")
```
nikvoloshin, line 66: I'd suggest use CI-related env variables. Usually `CI: true` env var is used.
And it'd better make `port` env var to, so we'll be able to change CI configuration without changing the code.

So, I'd suggest change this to:
`@EnabledIfEnvironmentVariable("CI", "true")`
`val port = System.getenv("PORT")`

nikvoloshin (changes requested):

nikvoloshin (approved):

PR #204, opened 2021-08-17
Implement mapdb context manager using Jackson for serialization

veptechno:
Now MapDbBotContextManager uses the serialization mechanism provided by java.io. This is inconvenient to use because it forces you to implement the Serializable interface for all classes that are stored in BotContext.

This PR adds a BotContextManager that uses MapDb and Jackson for serialization. 
**What's done**: created JacksonMapDbBotContextManager class, deprecated MapDbBotContextManager class, added tests for JacksonMapDbBotContextManager.

### DIFF
--- a/managers/mapdb/src/main/kotlin/com/justai/jaicf/context/manager/mapdb/BotContextModel.kt
+++ b/managers/mapdb/src/main/kotlin/com/justai/jaicf/context/manager/mapdb/BotContextModel.kt
@@ -4,21 +4,13 @@ import com.justai.jaicf.context.BotContext
 import com.justai.jaicf.context.DialogContext
 import java.io.Serializable
 
-internal class BotContextModel private constructor(): Serializable {
-
-    var result: Any? = null
-    lateinit var dialogContext: DialogContext
-    lateinit var client: Map<String, Any?>
-    lateinit var session: Map<String, Any?>
+class BotContextModel(botContext: BotContext): Serializable {
+    val dialogContext = botContext.dialogContext
+    val result: Any? = botContext.result
+    val client = botContext.client.toMap()
+    val session = botContext.session.toMap()
 
     companion object {
         private const val serialVersionUID = -8182528581552214215L
-
-        fun create(botContext: BotContext) = BotContextModel().apply {
-            dialogContext = botContext.dialogContext
-            result = botContext.result
-            client = botContext.client.toMap()
-            session = botContext.session.toMap()
-        }
     }
 }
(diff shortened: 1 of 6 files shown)


Denire (approved):

managers/mapdb/src/main/kotlin/com/justai/jaicf/context/manager/mapdb/BotContextModel.kt
```diff
@@ -1,15 +1,24 @@
...
 import com.justai.jaicf.context.BotContext
+import com.justai.jaicf.context.DialogContext
 import java.io.Serializable
 
-class BotContextModel(botContext: BotContext): Serializable {
-    val dialogContext = botContext.dialogContext
-    val result: Any? = botContext.result
-    val client = botContext.client.toMap()
-    val session = botContext.session.toMap()
+internal class BotContextModel private constructor(): Serializable {
```
nikvoloshin, line 7: Did you check this change with the old MapDB manager? As far as old MapDB works with java.io.Serializable we should double-check all changes in serializable classes

nikvoloshin: Also, we shouldn't change the visibility of public classes.
And anyway, I'd suggest remaining all old code unchanged and mark it deprecated.
And implement jackson-mapdb as completely new manager

managers/mapdb/src/main/kotlin/com/justai/jaicf/context/manager/mapdb/MapDbBotContextManager.kt
```diff
@@ -8,7 +8,14 @@ import com.justai.jaicf.context.manager.BotContextManager
 import org.mapdb.DBMaker
 import org.mapdb.Serializer
 
-class MapDbBotContextManager(dbFilePath: String? = null): BotContextManager {
+@Deprecated(
+    "MapDbBotContextManager is deprecated. Use JacksonMapDbBotContextManager instead",
+    ReplaceWith(
```
nikvoloshin, line 13: Let's not add replace with here. If a user will decide to use this quick fix, he will break his db-file

nikvoloshin: Also, let's add incompatibility notice in deprecation message

nikvoloshin (changes requested):

nikvoloshin (approved):

PR #206, opened 2021-08-22
 Actualize examples scenarios

veptechno:
**What's wrong**: The examples do not compile and use outdated reactions.
This PR fixes the examples to use relevant reactions and changes the state tree so the examples run on the current version of jaicf.

### DIFF
--- a/examples/game-clock/src/main/kotlin/com/justai/jaicf/examples/gameclock/GameController.kt
+++ b/examples/game-clock/src/main/kotlin/com/justai/jaicf/examples/gameclock/GameController.kt
@@ -62,7 +62,7 @@ class GameController(context: BotContext) {
 
     fun record(timeMs: Long) {
         currentTime = timeMs
-        overall = overall?.let { it.plus(timeMs) } ?: timeMs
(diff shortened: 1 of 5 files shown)


examples/game-clock/src/main/kotlin/com/justai/jaicf/examples/gameclock/GameController.kt
```diff
@@ -62,7 +62,7 @@ class GameController(context: BotContext) {
 
     fun record(timeMs: Long) {
         currentTime = timeMs
-        overall = overall?.let { it.plus(timeMs) } ?: timeMs
+        overall = overall?.plus(timeMs) ?: timeMs
```
nikvoloshin, line 65: Maybe something like that?
```suggestion
        overall = (overall ?: 0L) + timeMs
```

Or, we can even get rid of nullability using `withDefault`:
`var overall: Long by context.client.withDefault { 0L }`

nikvoloshin (approved):

Denire (approved):

PR #216, opened 2021-09-30
Add reactions rendering to ConsoleChannel

nikvoloshin:
This PR adds simple rendering of image, audio and buttons reaction for ConsoleChannel

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/channel/ConsoleChannel.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/channel/ConsoleChannel.kt
@@ -3,6 +3,7 @@ package com.justai.jaicf.channel
 import com.justai.jaicf.api.BotApi
 import com.justai.jaicf.api.QueryBotRequest
 import com.justai.jaicf.context.RequestContext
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/com/justai/jaicf/channel/ConsoleChannel.kt
```diff
@@ -15,41 +16,51 @@ import java.util.*
...
+
+        println() // flush conversation logs
+        println(answer)
+    }
+
+    private fun render(reaction: Reaction): List<String> = when (reaction) {
+        is SayReaction -> reaction.text.split("\n")
+        is ImageReaction -> listOf("${reaction.imageUrl} (image)")
+        is AudioReaction -> listOf("${reaction.audioUrl} (audio)")
+        is ButtonsReaction -> listOf(reaction.buttons.joinToString(" ") { "[$it]" })
```
Denire, line 63: What about other reactions like "go", "changeState", "video" and "file"? 
I think we should support all the reactions provided in core.

PR #217, opened 2021-10-01
Add CailaEntityActivationRule

nikvoloshin:
This PR brings support of caila entities by `CailaIntentActivator` and `CailaEntityActivationRule`


Denire:
@nikvoloshin Is it possible for caila package to add extensions for tests? For example:
```kotlin
class Test : ScenarioTest(){

  fun myTest(){
    cailaEntity("hello") goesToState "/Hello"
  }

}

```

activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/Duckling.kt
```diff
@@ -0,0 +1,12 @@
+package com.justai.jaicf.activator.caila
```
Denire, line 1: I'm not sure that we have to specify CAILA system entities in our sources. It's a nice to have object, but it may mislead user into thinking that CAILA can use only these entities, which is not true.

nikvoloshin: I can remove this, but there are many such objects already in JAICF.
Like `TelegramEvent`, `DialogflowIntent`, etc.

PR #225, opened 2021-11-24
Fix bug with transitionHistory. Rewrite managers' tests.

nikvoloshin:


core/src/main/kotlin/com/justai/jaicf/context/manager/InMemoryBotContextManager.kt
```diff
@@ -51,11 +51,26 @@ object InMemoryBotContextManager : BotContextManager {
...
+
     dc.nextContext = nextContext
     dc.currentContext = currentContext
     dc.nextState = nextState
     dc.currentState = currentState
-    dc.transitions.putAll(transitions)
-    dc.backStateStack.addAll(backStateStack)
+
+    dc.transitions.apply {
+        clear()
```
morfeusys, line 61: Sorry, didn't get that. Why do you need to clear an empty map of newly created `DialogContext`?

core/src/testFixtures/kotlin/com/justai/jaicf/core/test/managers/BotContextManagerBaseTest.kt
```diff
@@ -0,0 +1,108 @@
...
+    fun `Saves simple value`() {
+        val context = BotContext("client1").apply {
+            result = "some result"
+            session["key1"] = "some value"
+            client["key1"] = "some value"
+        }
+
+        val result = manager.exchangeContext(context)
+
+        Assertions.assertNotNull(result)
```
morfeusys, line 27: Why not to import `Assertions` methods?

morfeusys (approved):

core/src/main/kotlin/com/justai/jaicf/context/manager/InMemoryBotContextManager.kt
nikvoloshin: It's not necessary for `transitions` and `backStateStack`, but it's important for `transitionHistory` as it's created not empty.
And as far it's necessary for some of the fields, I think it's better to keep this logic for all fields in case of changes in implementation details.

core/src/testFixtures/kotlin/com/justai/jaicf/core/test/managers/BotContextManagerBaseTest.kt
nikvoloshin: Copy-pasted this from mongo-test. Fixed

veptechno (approved):

Denire (approved):

PR #226, opened 2021-12-08
polling: add stopPolling method, allow using provided executor, unify threads for http and execution

Denire:
There was no way to implement graceful shutdown with long polling and previous thread distribution made it easy to create a bottleneck in `ThreadPoolRequestExecutor`, as separate coroutines filled pool with requests and never ensured pool capacity.

Changes:
1. Creates jaicpConnector.stop() method to stop polling coroutines.
2. Allows providing Executor to run requests on.
3. Unifies thread pool for HTTP calls and request processing

### DIFF
--- a/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/JaicpConnector.kt
+++ b/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/JaicpConnector.kt
@@ -5,11 +5,13 @@ import com.justai.jaicf.channel.jaicp.channels.JaicpNativeChannelFactory
 import com.justai.jaicf.channel.jaicp.dto.ChannelConfig
 import com.justai.jaicf.channel.jaicp.dto.JaicpBotRequest
 import com.justai.jaicf.channel.jaicp.dto.JaicpResponse
-import com.justai.jaicf.channel.jaicp.execution.ThreadPoolRequestExecutor
+import com.justai.jaicf.channel.jaicp.execution.JaicpRequestExecutor
 import com.justai.jaicf.channel.jaicp.http.ChatAdapterConnector
 import com.justai.jaicf.helpers.http.toUrl
 import com.justai.jaicf.helpers.logging.WithLogger
 import io.ktor.client.*
+import java.util.concurrent.Executor
+import java.util.concurrent.Executors
 
 const val DEFAULT_PROXY_URL = "https://bot.jaicp.com"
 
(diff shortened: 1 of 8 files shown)


channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/polling/PollingChannel.kt
```diff
@@ -5,5 +5,18 @@ import com.justai.jaicf.channel.jaicp.JaicpBotChannel
 internal data class PollingChannel(
     val url: String,
     val botChannel: JaicpBotChannel,
-    val poller: RequestPoller
-)
\ No newline at end of file
+    val poller: RequestPoller,
+) {
```
nikvoloshin, line 9: I think we can get rid of isActive/start/stop logic in PollingChannel and move it to Dispatcher where it's used to be in order to simplify code flow a little bit

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/polling/RequestPoller.kt
```diff
@@ -25,16 +27,17 @@ internal class RequestPoller(
     }
 
     private var unprocessed: Boolean = false
+    private var isActive: Boolean = false
```
nikvoloshin, line 30: Shouldn't it be @Volatile in order to ensure the coroutine inside getUpdates always see right value?

Denire: Volatiles don't work for coroutines correctly. We can use something like `atomicfu`, but I think we don't need it.
These classes are internal and they have single point of interaction (dispatcher). I see no example use case when first thread calls run() and other calls stop() at the same time.

nikvoloshin: What about just AtomicBoolean from java?

Well, yes, it isn't a big problem even if the coroutine will see the updated value a little bit lately, but, anyway, it easy to fix with just AtomicBoolean

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/JaicpConnector.kt
```diff
@@ -31,10 +33,10 @@ abstract class JaicpConnector(
     val accessToken: String,
     val url: String,
     httpClient: HttpClient,
-    executorThreadPoolSize: Int = DEFAULT_REQUEST_EXECUTOR_THREAD_POOL_SIZE,
+    executor: Executor
```
nikvoloshin, line 36: Let's make this backward-compatible, as constructors of connectors are used widely by users.
Let's just make the new constructor primary and the old one secondary.
The same remark is for JaicpPollingConnector and JaicpWebhookConnector.

nikvoloshin (changes requested):
Everything is fine,  just make connectors backward compatible

nikvoloshin (changes requested):

nikvoloshin (approved):

veptechno (approved):

PR #228, opened 2022-01-12
Fix Rasa model classes

### LINKED ISSUE #227: Rasa NLU response parsing error
Hi

When trying to integrate RASA NLU, an exception is thrown when the NLU training set contains entities that are resolved. 

(From the console):
jaicf.activator.rasa.api.RasaApi - Cannot parse RasaParseMessageRequest(text=....., messageId=30eed22e-e5e8-4960-a979-6485d5295f45)
kotlinx.serialization.MissingFieldException: Field 'confidence' is required, but it was missing

From RASA output (using rasa train shell), we see a structure like this:

```
{
      "entity": "frequency",
      "start": 9,
      "end": 18,
      "confidence_entity": 0.9578643441200256,
      "value": "7",
      "extractor": "DIETClassifier",
      "processors": [
        "EntitySynonymMapper"
      ]
    }

```

But the mapping in RasaParseMessageResponse.kt contains:
```
@Serializable
data class Entity(
    val start: Int,
    val end: Int,
    val confidence: Float,
    val value: String,
    val entity: String
)
```

Maybe the confidence field should be renamed 'confidence_entity' to map correctly the output?

nikvoloshin:
Fix RASA NLU model classes in order to conform with [API spec](https://rasa.com/docs/rasa/pages/http-api#operation/parseModelMessage)

Fixes #227

### DIFF
--- a/activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/JSON.kt
+++ b/activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/JSON.kt
@@ -2,4 +2,4 @@ package com.justai.jaicf.activator.caila
 
 import kotlinx.serialization.json.Json
 
-internal val JSON = Json { ignoreUnknownKeys = true; encodeDefaults = true }
+internal val JSON = Json { ignoreUnknownKeys = true; isLenient = true; encodeDefaults = true }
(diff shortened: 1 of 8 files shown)


Denire:
I think we also should set `isLenient` on kotlinx.serializator for RASA (and maybe other activators or channels using KXS). 
```kotlin
serializer = KotlinxSerializer(Json { ignoreUnknownKeys = true; isLenient = true })
```
This will make serialization less strict

Denire (approved):

veptechno (approved):

PR #239, opened 2022-07-13
add modal and noContext fields to transition for activation selectors

Denire:
This PR adds more fields to `Transition` object. 

**Why?**
We need `modal` field to implement JAICP-like `ActivationSelector` in some projects.


Denire:
declined

PR #246, opened 2023-02-10
CailaIntentActivator - separate thresolds for patterns and phrases match

dimmarch:
CAILA uses different algorithms for pattern and phrases matching, so using same confidence threshold for both of them is not correct. After recent changes CAILA response contains weights for both of match types, so we can specify different thresholds in JAICF

Things i changed:
- added weights to `CailaInferenceResultData` 
- added optional thresholds for patterns and phrases in `CailaNLUSettings`
- added `CailaThresholdProcessor`:
-- it should apply current threshold settings to `CailaInferenceResultData`
-- Due to backward compatibility, if phrases or patterns threshold not provided, we should use `confidenceThreshold` field
-- overall result confidence can change after processor

### DIFF
--- a/activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaThresoldProcessor.kt
+++ b/activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaThresoldProcessor.kt
@@ -1,6 +1,5 @@
-package com.justai.zb.scenarios.engine
+package com.justai.jaicf.activator.caila
 
-import com.justai.jaicf.activator.caila.CailaNLUSettings
 import com.justai.jaicf.activator.caila.dto.CailaInferenceResultData
 import kotlin.math.max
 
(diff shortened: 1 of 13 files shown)


activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaThresoldProcessor.kt
```diff
@@ -0,0 +1,36 @@
+package com.justai.zb.scenarios.engine
```
nikvoloshin, line 1: Имя пакета не соответствует расположению файла

dimmarch: не заметил, спасибо. Поправил

nikvoloshin (approved):

activators/caila/src/main/kotlin/com/justai/jaicf/activator/caila/CailaThresoldProcessor.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.justai.jaicf.activator.caila.dto.CailaInferenceResultData
+import kotlin.math.max
+
+class CailaThresholdProcessor(
+    settings: CailaNLUSettings
+) {
+    private val patternThreshold = settings.intentThresholds.patterns ?: settings.confidenceThreshold
+    private val phrasesThreshold = settings.intentThresholds.phrases ?: settings.confidenceThreshold
+
+    //  Due to different thresholds for patters and phrases result.confidence can be changed after this method
```
Denire, line 12: It's better to write comments before method in javadoc format, not regular comments

Denire (approved):

PR #249, opened 2023-02-20
sendSms reaction

eshchkate:
This PR adds functionality for sending SMS messages from the phone channel using the JAICP platform.
The function is added following the analogy of existing reactions.

### DIFF
--- a/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/reactions/JaicpReactions.kt
+++ b/channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/reactions/JaicpReactions.kt
@@ -1,7 +1,12 @@
 package com.justai.jaicf.channel.jaicp.reactions
 
 import com.justai.jaicf.channel.jaicp.JSON
-import com.justai.jaicf.channel.jaicp.dto.*
+import com.justai.jaicf.channel.jaicp.dto.JaicpDialerAPI
+import com.justai.jaicf.channel.jaicp.dto.JaicpResponseData
+import com.justai.jaicf.channel.jaicp.dto.Reply
+import com.justai.jaicf.channel.jaicp.dto.TextReply
+import com.justai.jaicf.channel.jaicp.dto.ProviderConfig
+import com.justai.jaicf.channel.jaicp.dto.SmsReply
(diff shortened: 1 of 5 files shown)


channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/reactions/JaicpReactions.kt
```diff
@@ -45,6 +48,15 @@ open class JaicpReactions : Reactions() {
         return SayReaction.create(text)
     }
 
+    /**
+     * Send SMS from scenario. You can send SMS messages from the phone channel.
+     * To do this, the SMS package must be connected to the JAICP platform.
+     * */
+    fun sendSms(text: String, phoneNumber: String, provider: ProviderConfig? = null): SmsReaction {
```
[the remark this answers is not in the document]
eshchkate: no, only jaicp-native channel

channels/jaicp/src/main/kotlin/com/justai/jaicf/channel/jaicp/reactions/reaction/SmsReaction.kt
```diff
@@ -0,0 +1,21 @@
+package com.justai.jaicf.channel.jaicp.reactions.reaction
+
+import com.justai.jaicf.channel.jaicp.dto.ProviderConfig
+import com.justai.jaicf.channel.jaicp.dto.SmsReply
+import com.justai.jaicf.logging.Reaction
+
+data class SmsReaction(
+    val text: String? = null,
+    val phoneNumber: String? = null,
+    val provider: ProviderConfig? = null,
```
Denire, line 10: I'm not sure that this providerConfig should be saved to reaction as it contains password and sensitive data. I guess it's better to remove it.

eshchkate: We need it to set up the sender name (if desired). Mb let's change it to a JsonObject?

Denire:
I think we should discard this PR and make SMS available to all channels. This requires some changes in JAICP components for framework to use. 
Final architecture will look somewhat like JAICP Livechats, where any channel implementing given specific interface with receive an extension method for SMS. This method will be working only if channel is connected via `JaicpConnector`.

PR #250, opened 2023-02-24
support only-if activation predicates on AnyIntent/AnyEvent

Denire:
Adds `anyIntent` or `anyEvent` activation predicates support for onlyIf activation rules.

**What was the problem:**
postProcessing for `anyIntent` and `anyEvent` didn't copy onlyIf predicates recreating ActivationRules with new matching method.
```
transitions.map {
    when (it.rule) {
        is AnyIntentActivationRule -> it.copy(rule = AnyIntentActivationRule(intents))
        is AnyEventActivationRule -> it.copy(rule = AnyEventActivationRule(events))
        else -> it
    }
}
```

**What's changed**
We start copying onlyIf predicates to new ActivationRules in `postProcess`

### DIFF
--- a/core/src/main/kotlin/com/justai/jaicf/model/ScenarioModelBuilder.kt
+++ b/core/src/main/kotlin/com/justai/jaicf/model/ScenarioModelBuilder.kt
@@ -65,20 +65,26 @@ internal class ScenarioModelBuilder {
     }
 
     private fun postProcess(): ScenarioModel {
-        val newTransitions = transitions.groupBy { it.fromState }.flatMap { (_, transitions) ->
+        transitions.groupBy { it.fromState }.forEach { (_, transitions) ->
             val intents = transitions.map { it.rule }.filterIsInstance<IntentByNameActivationRule>().map { it.intent }
             val events = transitions.map { it.rule }.filterIsInstance<EventByNameActivationRule>().map { it.event }
 
-            transitions.map {
+            transitions.forEach {
                 when (it.rule) {
-                    is AnyIntentActivationRule -> it.copy(rule = AnyIntentActivationRule(intents))
-                    is AnyEventActivationRule -> it.copy(rule = AnyEventActivationRule(events))
-                    else -> it
+                    is AnyIntentActivationRule -> it.rule.except.apply {
+                        clear()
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/justai/jaicf/model/activation/ActivationRule.kt
```diff
@@ -50,4 +50,9 @@ abstract class ActivationRuleAdapter : ActivationRule {
         onlyIf.add(predicate)
         return this
     }
+
+    fun getOnlyIfPredicates() = onlyIf
+
+    fun setOnlyIfPredicates(predicates: List<ActivationRule.OnlyIfContext.() -> Boolean>) =
+        onlyIf.addAll(predicates)
```
nikvoloshin, line 57: Using `add` function in a `set` method might be confusing. I'd suggest to use a single verb and logic in a method's name and method's code. Either leave it as `set`, but change the code to `onlyIf = predicates`, or rename the method to `add`.
If you choose `set` logic, it might be implemented as a simple `var` property

core/src/main/kotlin/com/justai/jaicf/model/ScenarioModelBuilder.kt
```diff
@@ -102,4 +109,8 @@ internal class ScenarioModelBuilder {
         }
         return ScenarioModel(newStates.associateBy { it.path.toString() }, newTransitions, newHooks)
     }
-}
\ No newline at end of file
+}
+
+private fun ActivationRuleAdapter.copyOnlyIfPredicates(other: ActivationRuleAdapter) = apply {
```
nikvoloshin, line 114: Should the mutation be that way around?
Instead of making `onlyIf` a mutable property for every `ActivationRule`, we can simply make exclusions of `any*` rules mutable and leave base class as is.
Another reason for don't making mutation on a base class level is that now we have to support a mutation of all base class properties in order to set them on `any` rules processing

Denire: that's the right idea, thanks! i've updated this PR

nikvoloshin (approved):

